PR #13, opened 2021-08-18
Add example on dataset distillation

fabianp:


examples/dataset_distillation.py
```diff
@@ -0,0 +1,83 @@
+# -*- coding: utf-8 -*-
```
mblondel, line 1: Add a license header.

examples/dataset_distillation.py
```diff
@@ -0,0 +1,83 @@
...
+
+inner_loss = objective.l2_multiclass_logreg
+gd = GradientDescent(fun=inner_loss, tol=1e-3, maxiter=500)
+
+def outer_loss(img):
+    # inner_sol is the solution to the inner problem, which computes the
+    # model trained on the 10 images from distilled_images. This makes
+    # the problem bi-level, as the objective depends itself on the solution
+    # of an optimization problem  (inner_sol)
+    inner_sol, _ = gd.run(params, l2reg, (img, distilled_labels))
```
mblondel, line 67: Since the output is a named tuple, you could also write it `inner_sol = gd.run(params, l2reg, (img, distilled_labels)).params` but the way you write it is fine too.

mblondel (commented):
Thanks a lot Fabian. LGTM, apart from the license header.

fabianp:
Thanks for the review. I believe I've addressed the comments

Issue #16, opened 2021-08-23
Bisection hanging 


jjruby09:
I am trying to jaxopt.Bisection to replace the use of scipy.optimize.bisect in a computational model but Bisection hangs when I run my code. 

The basic structure includes 2 functions that are both jitted (so I assume it should be able to compile ok): 

```
@jit
def f1(parameters):
    ....
    return jax.numpy.array([a,b,c])

@jit
def opt_fun(x):
    f1(x,params)
    .... 
    return float_value
```

when I call `scipy.optimize.bisect(opt_fun,x0,x1)` it runs with no issue but `jaxopt.Bisection(opt_fun,x0,x1).run(None)` hangs with with~10% cpu usage and55% memory usage on i9 2018 macbook pro with 32GB of memory. 

I acknowledge I may be using this incorrectly and that this is possibly not the intended use case but any direction would be very helpful. My intention is to use this computational model with numpyro in the future and having a jax version of the bisection root finding would be incredibly helpful.

mblondel:
Thanks a lot for the report! Could you provide a short script for reproducing the issue?

jjruby09:
I made many attempts at creating a small script to reproduce the problem and unfortunately failed. I apologize in advanced that this is a mess to look at but the full script is the only place I was able to get the issue to happen (leading me to believe that it is likely something specific about my model but I am not very adept at diagnosing the compilations that occur). 

```
import numpy as np # 1.21.2
import jax.numpy as jnp # jax version = 0.2.19
from jax.experimental import ode
from jax import jit
from jax.config import config
config.update("jax_enable_x64", True)

#############################################################################
'''
All of the following is establishing constants and equations used throughout
'''
#############################################################################
sigma = 5.670374419E-8 # Stefan-Boltzman constant in W m^-2 K^-4
G = 6.67408E-11 #Gravitational constant in m^3 kg^-1 s^-2

#solar constants
Rsun = 696.34E6 #meters
Msun = 1.9891E30 #kg
Lsun = 3.846E26 #Watts

h = 6.62606957e-34 #Planck constant in  m^2 kg/s
hbar = 1.05457173e-34      # h/2pi
kb = 1.38064852E-23 #Boltzman constant in  m^2 kg / (s^2 K)
me = 9.109E-31 #mass of electron in  kg
mu = 1.66053904E-27 #atomic mass unit in kg
mp = 1.67262178e-27 # mass of proton

[...]

PR #19, opened 2021-08-27
Bisection improvements.

mblondel:
- Docstring improvements
- Set `init_params` to `None` by default
- Convert `lower` and `upper` to floats (see #16)

### DIFF
--- a/jaxopt/_src/bisection.py
+++ b/jaxopt/_src/bisection.py
@@ -35,6 +35,7 @@ class BisectionState(NamedTuple):
   error: float
   low: float
   high: float
+  sign: int
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


jaxopt/_src/bisection.py
```diff
@@ -118,7 +127,8 @@ def update(self,
     Returns:
       (params, state)
     """
-    value = self._fun(params, *args, **kwargs)
+    value = self.optimality_fun(params, *args, **kwargs)
+    value = jnp.where(state.sign == 1, value, -value)
     too_large = value > 0
```
shoyer, line 132: Maybe:
```suggestion
    too_large = state.sign * value > 0
```

mblondel:
@shoyer Addressed your comment, thanks!

shoyer (approved):

PR #24, opened 2021-09-03
Example gallery using sphinx-gallery

fabianp:


mblondel:
I just built the documentation and add to install  tensorflow (2.6.0) and tensorflow-datasets (4.4.0). 

We should probably add a `requirements_examples.txt` file and add in docs/README.md `pip install -r requirements_examples.txt`.

docs/index.rst
```diff
@@ -24,6 +24,7 @@ You may install JAXopt from GitHub with:
    stochastic
    implicit_diff
    root_finding
+   auto_examples/index
```
mblondel, line 27: Technically, examples are not part of the public API. So I would prefer

**Public API**

* Basics
* Unconstrained optimization
* Constrained optimization
* Non-smooth optimization
* Stochastic optimization
* Implicit differentiation
* Root finding

**Examples**

- ...
- ...
- ...

mblondel:
Awesome work! LGTM apart from my small remarks.

mblondel (approved):

docs/index.rst
fabianp: Screenshot-20210903175942-960x1056

How about the above? Almost what you want, but without expanding the examples. I think we should do this once we start putting the examples into different folders (as sklearn does).

mblondel: Why is it necessary to put examples in different folders?

fabianp: Because I don't know how to expand the examples otherwise. I tried changing the :maxdepth: in the toctree, like 
```
.. toctree::
   :maxdepth: 3
   :caption: Examples

   auto_examples/index
```

but it doesn't seem to change anything

mblondel: I see! I'm fine with the current solution for now then.

PR #28, opened 2021-09-10
Fix ridge regression example: change x0 to init in solve_cg

tachukao:
The current ridge regression example is broken with the following error:

```
TypeError: jax._src.scipy.sparse.linalg.cg() got multiple values for keyword argument 'x0'
```

I believe this stems from the fact that `linear_solve.solve_cg` takes `init` as opposed to `x0` as a named parameter

https://github.com/google/jaxopt/blob/e42a1ff6f80ca2a3dae5fa60c2a7fedad91d527c/jaxopt/_src/linear_solve.py#L109


tachukao:
@googlebot I signed it!

mblondel:
Good catch! Thanks a lot.

PR #29, opened 2021-09-15
Add implicit_diff_solve, jit, unroll options and reorganize base classes

### LINKED ISSUE #27: pre_update for OptaxSolver run function?
Hi,
Defining a pre_update function works well in OptaxSolver when using the run_iterator function (like in the MNIST example). However, it does not work with the run function. I checked the implementation and indeed there is no call to the pre_update inside '[run](https://github.com/google/jaxopt/blob/8fb5c2ca490620832d1cfd2e71eb641ce225c1f7/jaxopt/_src/base.py#L69)' but [there is one](https://github.com/google/jaxopt/blob/8fb5c2ca490620832d1cfd2e71eb641ce225c1f7/jaxopt/_src/base.py#L160) inside 'run_iterator'. Is this a small bug or is this by design?

Thanks!

mblondel:
This PR implements several core changes.

1) Add an `implicit_diff_solve` option for specifying the linear solver to use for doing implicit diff (previously the `implicit_diff` served *both* the purpose of enabling/disabling implicit diff and of specifying the solver)

2) Add `jit` and `unroll` option to all solvers for finer-grained control on how the optimization loop is performed

3) Move `pre_update` to within `update` in `OptaxSolver` and `PolyakSGD` (fixes #27)

4) Reorganize and rename classes. Since Python is increasingly used as a typed language, it will be useful if we can create functions that accept a solver as argument
```
def some_func(solver: jaxopt.base.IterativeSolver):
```

### DIFF
--- a/jaxopt/_src/bisection.py
+++ b/jaxopt/_src/bisection.py
@@ -39,7 +39,7 @@ class BisectionState(NamedTuple):
 
 
 @dataclass
-class Bisection:
+class Bisection(base.IterativeSolver):
   """One-dimensional root finding using bisection.
 
   Attributes:
@@ -52,9 +52,11 @@ class Bisection:
     tol: tolerance.
     check_bracket: whether to check correctness of the bracketing interval.
       If True, the method ``run`` cannot be jitted.
-    implicit_diff: if True, enable implicit differentiation.
+    implicit_diff_solve: the linear system solver to use.
     verbose: whether to print error on every iteration or not.
       Warning: verbose=True will automatically disable jit.
+    jit: whether to JIT-compile the bisection loop (default: "auto").
+    unroll: whether to unroll the bisection loop (default: "auto").
 
   """
   optimality_fun: Callable
@@ -63,8 +65,10 @@ class Bisection:
   maxiter: int = 30
   tol: float = 1e-5
   check_bracket: bool = True
-  implicit_diff: bool = True
+  implicit_diff_solve: Callable = linear_solve.solve_lu
   verbose: bool = False
+  jit: base.AutoOrBoolean = "auto"
+  unroll: base.AutoOrBoolean = "auto"
 
   def init(self,
            init_params=None,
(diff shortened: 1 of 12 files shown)


froystig (approved):
Looks good, especially the jit/unroll "auto" logic.

jaxopt/_src/bisection.py
```diff
@@ -63,8 +65,10 @@ class Bisection:
   maxiter: int = 30
   tol: float = 1e-5
   check_bracket: bool = True
-  implicit_diff: bool = True
```
fllinares, line 66: Naive question: shouldn't we keep `implicit_diff` as an attribute just in case we want to set it to False?

jaxopt/_src/scipy_wrappers.py
```diff
@@ -211,19 +211,16 @@ class ScipyWrapper(abc.ABC):
...
+    implicit_diff_solve: the linear system solver to use.
     has_aux: whether function `fun` outputs one (False) or more values (True).
       When True it will be assumed by default that `fun(...)[0]` is the
       objective.
   """
   method: Optional[str] = None
   dtype: Optional[Any] = onp.float64
   jit: bool = True
-  implicit_diff: Union[bool, Callable] = False
+  implicit_diff: bool = False
```
fllinares, line 222: Appears to be missing from docstring. Besides, since autodiff via unrolling isn't supported by the SciPy wrappers, should we just remove the attribute?

fllinares (approved):
Thanks a lot Mathieu, this looks great!! I just left a couple of minor comments / questions.

jaxopt/_src/scipy_wrappers.py
mblondel: When `run` is decorated with `@custom_root`, which happens if `implicit_diff=True`, keyword arguments are no longer usable in `run`. I wanted to remove `implicit_diff` here but this broke some of your tests using keyword arguments. So I'm leaving `implicit_diff` temporarily until we fix the keyword issue (@froystig and I think we have a fix).

jaxopt/_src/bisection.py
mblondel: I think unrolling will not work since bisection is just halving the bracketing interval on each iteration. So `implicit_diff=False` doesn't really make sense.

jaxopt/_src/scipy_wrappers.py
fllinares: Makes sense. Please let me know if you'd prefer me to change the tests to not rely on kwargs.

jaxopt/_src/bisection.py
fllinares: I see, also makes sense, thanks! 

Extremely minor, but should `base.IterativeSolver`'s docstring describe `implicit_diff` as a needed attribute then? The only other reason I could find to keep it is what you mentioned in the other comment about kwargs, but probably not worth considering anyway.

jaxopt/_src/scipy_wrappers.py
mblondel: No that's fine. @froystig and I will tackle the issue soon.

jaxopt/_src/bisection.py
mblondel: Done.

PR #32, opened 2021-09-17
Quad solver fix

gowerrobert:
Changed the linear solver used  in _solve_eq_constrained_qp_KKT to gminres since the KKT is symmetric but not necessarily positive definite.  This is the recommend solver for this  type of KKT system in Nocedal & Wright, section 16.3, Numerical Optimization. 

I also tried solve_normal_cg instead, but it's convergence in residual is slow, and consequently fails the units test _check_derivative_A_and_b unless significantly increase tolerance (and number of iterations).

### DIFF
--- a/jaxopt/_src/linear_solve.py
+++ b/jaxopt/_src/linear_solve.py
@@ -154,6 +154,7 @@ def solve_normal_cg(matvec: Callable,
 def solve_gmres(matvec: Callable,
                 b: Any,
                 ridge: Optional[float] = None,
+                tol: float = 1e-5,
                 **kwargs) -> Any:
   """Solves ``A x = b`` using gmres.
 
@@ -168,7 +169,7 @@ def solve_gmres(matvec: Callable,
   """
   if ridge is not None:
     matvec = _make_ridge_matvec(matvec, ridge=ridge)
-  return jax.scipy.sparse.linalg.gmres(matvec, b, **kwargs)[0]
+  return jax.scipy.sparse.linalg.gmres(matvec, b, tol=tol, **kwargs)[0]
 
 
 def solve_bicgstab(matvec: Callable,
--- a/jaxopt/_src/quadratic_prog.py
+++ b/jaxopt/_src/quadratic_prog.py
@@ -69,7 +69,9 @@ def _solve_eq_constrained_qp(init_params,
                              c,
                              matvec_A,
                              b,
-                             maxiter):
+                             maxiter,
+                             tol=1e-5,
+                             solve=linear_solve.solve_gmres): 
   """Solves 0.5 * x^T Q x + c^T x subject to Ax = b.
 
   This solver returns both the primal solution (x) and the dual solution.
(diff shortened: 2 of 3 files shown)


gowerrobert:
@googlebot I signed it!

jaxopt/_src/linear_solve.py
```diff
@@ -154,6 +154,7 @@ def _matvec(x):
 def solve_gmres(matvec: Callable,
                 b: Any,
                 ridge: Optional[float] = None,
+                tol: float = 1e-05,
```
mblondel, line 157: `1e-5`

jaxopt/_src/quadratic_prog.py
```diff
@@ -85,8 +86,7 @@ def matvec(u):
   # Solves the following linear system:
   # [[Q A^T]  [primal_var = [-c
   #  [A 0  ]]  dual_var  ]    b]
-  return linear_solve.solve_cg(matvec, (minus_c, b), init=init_params,
-                               maxiter=maxiter)
+  return solve(matvec, (minus_c, b),tol=1e-07, maxiter=maxiter)
```
mblondel, line 89: Add a `tol` option to the function instead and use `1e-5` as default  for consistency. Add a space before the comma.

jaxopt/_src/quadratic_prog.py
```diff
@@ -177,7 +177,8 @@ def run(self,
           init_params: Optional[Tuple] = None,
           params_obj: Optional[ArrayPair] = None,
           params_eq: Optional[ArrayPair] = None,
-          params_ineq: Optional[ArrayPair] = None) -> base.OptStep:
+          params_ineq: Optional[ArrayPair] = None,
+          ) -> base.OptStep:
```
mblondel, line 181: Any good  reason for this change? If not, I would change back for consistency with the rest of the code base.

mblondel (changes requested):
Thanks a lot Robert for the very good catch, and also for testing gmres vs normal_cg.

I left some cosmetic comments.

mblondel:
Another request, could you squash the commits into a single commit and force push after that? Thanks!

for squashing: `$ git rebase -i main`
for force push: `$ git push origin quad_solver_fix --force`

jaxopt/_src/quadratic_prog.py
gowerrobert: Changed to 1e-5 in linear_solve.solve_gmres instead for consistency. Otherwise doesn't pass _check_derivative_A_and_b test

mblondel: I see! I think the right way to do it is to add `tol` as an attribute of `QuadraticProgramming` and then you can call `_solve_eq_constrained_qp(..., tol=self.tol)`. Then, in your test, you can instantiate `QuadraticProgramming(..., tol=1e-7)`.

gowerrobert: ok, done that!

gowerrobert:
Did the commit squashing work as expected?

mblondel:
> Did the commit squashing work as expected?

Yes, thanks a lot!

The only change left is `solve_tol` -> `tol` for consistency with other solvers (all iterative solvers have a `tol` argument).

mblondel (approved):
Thanks Robert!

PR #34, opened 2021-09-20
loosen constraints on solver signatures

froystig:
When defining implicit derivatives, allow solver functions to have whatever signature, including keyword arguments (whether explicit or via a `**kwargs` binder). If differentiated, enforce that the call is consistent with the signature of the optimality function or the fixed-point function, for `custom_root` and `custom_fixed_point` derivatives respectively.

### DIFF
--- a/jaxopt/_src/implicit_diff.py
+++ b/jaxopt/_src/implicit_diff.py
@@ -14,7 +14,7 @@
 
 """Implicit differentiation of roots and fixed points."""
 
-
+import inspect
 from typing import Any
 from typing import Callable
 from typing import Tuple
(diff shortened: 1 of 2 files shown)


jaxopt/_src/implicit_diff.py
```diff
@@ -120,29 +120,124 @@ def root_jvp(optimality_fun: Callable,
...
+
+      ba_args, ba_kwargs, map_back = _signature_bind_and_match(
+          optimality_signature, *args, **kwargs)
+      if ba_kwargs:
+        raise TypeError(
+            "keyword arguments to solver_fun could not be resolved to "
+            f"positional arguments based on the signature {signature}. This "
+            "can happen under custom_root if optimality_fun takes catch-all "
+            "**kwargs, or under custom_fixed_point if fixed_point_fun takes "
+            "catch-all **kwargs, both of which are currently unsupported.")
```
froystig, line 217: This error message is rather specific. We may want to catch and re-raise it at higher levels of the library with a more helpful message.

jaxopt/_src/implicit_diff.py
```diff
@@ -120,29 +120,124 @@ def root_jvp(optimality_fun: Callable,
   return solve(matvec, Jv)
 
 
-def _custom_root(solver_fun, optimality_fun, solve, has_aux):
-  def solver_fun_fwd(init_params, *args):
-    res = solver_fun(init_params, *args)
-    return res, (res, args)
+def _extract_kwargs(kwarg_keys, args):
```
mblondel, line 123: Maybe rename `args` above to `flat_args` (which is consistent with below) or `all_args`. I wasn't sure in the beginning if `args` referred to positional-only arguments or to all arguments.

mblondel (approved):
Great work, thanks @froystig!

jaxopt/_src/implicit_diff.py
froystig: Good call. Done.

PR #37, opened 2021-09-23
Switch back to implicit_diff=True by default.

mblondel:


fllinares (approved):
Thanks Mathieu, all LGTM!

PR #41, opened 2021-09-24
Improve consistency between math and code.

mblondel:


fabianp (approved):
good changes, definitely an improvement in the doc. thanks!

PR #46, opened 2021-09-28
fix for tree_vdot for when leaves are not arrays. 

### LINKED ISSUE #44: tree_util.tree_vdot bugging on example.
https://github.com/google/jaxopt/blob/1d553785f1755249d0e546c174619a0e70e17b7e/jaxopt/_src/tree_util.py#L51

See [this colab](https://colab.research.google.com/drive/1m_eYcxYvCbjVOiHbOHDHzB-KuqBknzHD#scrollTo=4DZ3YA_T2tkz) for details on bug and fix. 

Here's a minimal example to reproduce the bug:

```python
from jaxopt import tree_util
a =(1.0, {"k1": 1.0, "k2": (1.0, 1.0)}, 1.0)
x =(1.0, {"k1": 2.0, "k2": (3.0, 4.0)}, 5.0)
out = tree_util.tree_vdot(a,x)
```

The issue is caused by unexpected behaviour of ```Jax.tree_util.tree_multimap```.  

One way to fix this is to broadcast a product with tree_multimap, and then reduce using the sum
```python

def tree_vdot(tree_x, tree_y):
  """Compute the inner product <tree_x, tree_y>."""
  prod_pair = tu.tree_multimap(lambda x, y: x*y, tree_x, tree_y)
  sums = tu.tree_map(jnp.sum, prod_pair)
  return tu.tree_reduce(operator.add, sums)
```
 I would be happy to implement this fix, implement new accompanying unit test, and make a pull request. More details here

gowerrobert:
Fixes #44

1. Added a _vdot_safe product for computing tree_vdot (Mathieu's fix)
2. Added new test on pytrees that do not have arrays on leaves

### DIFF
--- a/jaxopt/_src/tree_util.py
+++ b/jaxopt/_src/tree_util.py
@@ -48,9 +48,12 @@ def tree_add_scalar_mul(tree_x, scalar, tree_y):
 _vdot = functools.partial(jnp.vdot, precision=jax.lax.Precision.HIGHEST)
 
 
+def _vdot_safe(a, b):
+  return _vdot(jnp.asarray(a), jnp.asarray(b))
+
 def tree_vdot(tree_x, tree_y):
   """Compute the inner product <tree_x, tree_y>."""
-  vdots = tree_multimap(_vdot, tree_x, tree_y)
+  vdots = tree_multimap(_vdot_safe, tree_x, tree_y)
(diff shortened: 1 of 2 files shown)


mblondel (approved):
LGTM, thanks a lot!

PR #55, opened 2021-10-02
Add AndersonWrapper

Algue-Rythme:
I added AndersonWrapper class to accelerate Jaxopt optimizers with Anderson acceleration.
Several modifications have been made:

- AndersonWrapper and its tests
- AndersonWrapper documentation
- AndersonWrapper example on Block CD
- AndersonAcceleration refactorization
- pytree are supported in history => no need for tree_collapse or tree_uncollapse
- scalar are now supported in Anderson (not only arrays)

Issue introduced:
- as result of pytree usage, the linear system is slower to solve: building GTG is a bit more expansive (see TODO).

By default I put this feature in V0.1 (in changelog) but I can move it to 0.2 if necessary.

Todo list (in future PR ?):
- correct speed issue
- add option for modulo in Anderson


Algue-Rythme:
I forced push some modifications to solve the speed issue.

Speed in main for anderson_tests: 17s

Speed before correction: 31s
Speed after correction (current state of my PR): 21s

Algue-Rythme:
After further reflexion, it appears that a single pytree with batched leafs is more Jax-like than a list of pytrees with unbatched leafs. It allows to benefit from hardware acceleration much more easily when `params` is big. This new push reflects this change in history.

jaxopt/_src/tree_util.py
```diff
@@ -78,38 +78,35 @@ def tree_zeros_like(tree_x):
...
-def tree_uncollapse(tree, vec):
-  """Returns a pytree with same elements as 1D vector ``vec`` and with same structure as ``tree``.
+def pytree_at_set(tree, pos, values):
+  at_set = lambda x, value: x.at[pos].set(value)
+  return tree_map(at_set, tree, values)
 
-  Reciprocal function of ``tree_collapse``. Transform the 1D vector into a pytree
-  compatible with ``tree`` structure, with same shapes.
+
+def tree_batch_vdot(a, b):
```
mblondel, line 99: Doesn't seem to be used anywhere -> delete.

tests/anderson_wrapper_test.py
```diff
@@ -0,0 +1,170 @@
...
+    lam = 10.0
+    w_init = jnp.zeros(data_train[0].shape[1])
+    tol = 1e-5
+    maxiter = 1000  # large number of updates
+    pg = ProximalGradient(fun=fun, prox=prox.prox_lasso, maxiter=maxiter, tol=tol,
+                          acceleration=False)
+    aw = AndersonWrapper(pg, history_size=5)
+    data_val = datasets.make_regression(n_samples=100, n_features=3, random_state=0)
+    def solve_run(lam):
+      aw_params, _ = aw.run(w_init, lam, data_train)
```
mblondel, line 162:
```suggestion
      aw_params = aw.run(w_init, lam, data_train).params
```

tests/anderson_test.py
```diff
@@ -114,8 +125,7 @@ def f(x, r, y):
     sol, state = aa.init(x0, r, y0)
     sols = []
     for i in range(10):
-      _, y = state.value
-      sol, state = aa.update(sol, state, r, y)
+      sol, state = aa.update(sol, state, r, state.aux)
```
mblondel, line 128: I noticed a similar test for `FixedPointIteration` and was a bit confused by the fact that `state.aux` is used to feed extra arguments to the function. This means that it wouldn't be possible to do the equivalent with `run`. Normally, the extra arguments are supposed to be constant. And the argument we iterate over is supposed to be the first one.

jaxopt/_src/tree_util.py
Algue-Rythme: I use it to build the Gram matrix in history. I renamed it to reflect its true semantic.

examples/fixed_point/plot_anderson_accelerate_gd.py
```diff
@@ -102,8 +102,8 @@ def T(params, eta, l2reg, data):
 # Plot error as function of iteration num
 ax = fig.add_subplot(spec[2, :])
 iters = jnp.arange(len(aa_errors))
-ax.plot(iters, aa_errors, label='Anderson Accelerated GD Error')
 ax.plot(iters, fp_errors, label='Gradient Descent Error')
+ax.plot(iters, aa_errors, label='Anderson Accelerated GD Error')
```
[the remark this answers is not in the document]
Algue-Rythme: This is already Gradient Descent without acceleration. This example is maybe a bit misleading: it accelerates an hard-coded gradient descent (with constant step size) using AndersonAcceleration. It is not an example of AndersonWrapper usage.

tests/anderson_test.py
Algue-Rythme: I understand. The example has been modified.

jaxopt/_src/anderson_wrapper.py
```diff
@@ -0,0 +1,149 @@
...
+    """
+    m = self.history_size
+    p_h = aa_state.params_history
+    r_h = aa_state.residuals_history
+    GTG = aa_state.GTG
+    pos = jnp.mod(aa_state.iter_num, self.history_size)
+
+    aa_params = anderson_step(p_h, r_h, GTG, self.ridge, self.beta)
+    solver_state = self.get_solver_state(aa_params, aa_state, *args, **kwargs)
+    faa_params, solver_state = self.solver.update(aa_params, solver_state, *args, **kwargs)
```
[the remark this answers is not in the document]
mblondel: Makes sense. I wonder if a variant with a `alpha >=0, sum(alpha) = 1` would work. This way we would stay in the feasible set.

PR #56, opened 2021-10-05
take optional reference signature for implicit diff

### LINKED ISSUE #51: name 'signature' is not defined
I get the following error "JaxStackTraceBeforeTransformation: NameError: name 'signature' is not defined" 
 with the following example when I set `implicit_diff=True`, but it works when I set `implicit_diff=False`

```
import jaxopt
import jax 
import jax.numpy as jnp 
import matplotlib.pyplot as plt 
jax.config.update("jax_enable_x64", True)

X = jax.random.normal(jax.random.PRNGKey(1), (100, 10))
y =  jax.random.normal(jax.random.PRNGKey(2), (100, 1))

def ridge_reg_objective(params, l2reg, X, y):
  residuals = jnp.dot(X, params) - y
  return jnp.mean(residuals ** 2) + l2reg*jnp.linalg.norm(params)

def ridge_reg_solution(l2reg, X, y):
  gd = jaxopt.GradientDescent(fun=ridge_reg_objective, maxiter=500, implicit_diff=False)
  return gd.run(init_params, l2reg=l2reg, X=X, y=y).params

l2reg = 0.01
init_params = jax.random.normal(jax.random.PRNGKey(0), shape=(X.shape[1],))
print(jax.jacobian(ridge_reg_solution, argnums=0)(l2reg, X, y))
```

froystig:
Instead of always relying on `optimality_fun` to carry the canonical signature for the solver, accept an optional canonical-signature function in `custom_root` and `custom_fixed_point`.

Check for a corresponding optional attribute on subclasses of `IterativeSolver` when run under implicit diff mode. Use this to indicate that `GradientDescent.fun` has the canonical signature for GD solvers.

Fixes #51

### DIFF
--- a/jaxopt/_src/implicit_diff.py
+++ b/jaxopt/_src/implicit_diff.py
@@ -17,6 +17,7 @@
 import inspect
 from typing import Any
 from typing import Callable
+from typing import Optional
 from typing import Tuple
 
 import jax
@@ -167,7 +168,8 @@ def _signature_bind_and_match(signature, *args, **kwargs):
   return out_args, out_kwargs, map_back
 
 
-def _custom_root(solver_fun, optimality_fun, solve, has_aux):
+def _custom_root(solver_fun, optimality_fun, solve, has_aux,
+                 reference_signature_fun=None):
   # When caling through `jax.custom_vjp`, jax attempts to resolve all
   # arguments passed by keyword to positions (this is in order to
   # match against a `nondiff_argnums` parameter that we do not use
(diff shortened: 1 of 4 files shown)


jaxopt/_src/implicit_diff.py
```diff
@@ -253,23 +261,30 @@ def custom_root(optimality_fun: Callable,
       solution / root ``sol``.
     has_aux: whether the decorated solver function returns auxiliary data.
     solve: a linear solver of the form ``solve(matvec, b)``.
+    reference_signature_fun: optional function whose signature
+      (i.e. arguments and keyword arguments) is one with which the
+      solver and optimality functions are expected to agree. Defaults
+      to ``optimality_fun``.
```
mblondel, line 267: Maybe briefly explain why this is needed / useful.

mblondel (approved):
Thanks a lot @froystig!

PR #60, opened 2021-10-07
Add modulo option, simplify init

Algue-Rythme:
Add Modulo option for both AndersonAcceleration and AndersonWrapper. Simplify their `init()` method by pushing the first iterates in `update()`.

We observe that using `modulo=history_size` instead of `modulo=1` does not decrease dramatically the convergence speed (see examples).


Algue-Rythme:
I added the examples of Anderson in the changelog.

PR #62, opened 2021-10-08
Add SGD with Armijo line search

Algue-Rythme:
Add support for Armijo Line Search. Different options for stepsise resetting are available, such as Goldstein Line search.  Add corresponding tests and documentation.

Rename `step_size` into `stepsize` in PolyakSGD for API consistency with `ProximalGradientDescent`.

Add an example in section `stochastic` to compare different Gradient Descent algorithms on simple classification task (under interpolation hypothesis), namely Gradient Descent, Proximal Gradient Descent, PolyakSGD and ArmijoSGD.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -39,6 +39,7 @@ Stochastic
 .. autosummary::
   :toctree: _autosummary
 
+    jaxopt.ArmijoSGD
     jaxopt.OptaxSolver
     jaxopt.PolyakSGD
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+Version 0.1.1 (development version)
+-----------------------------------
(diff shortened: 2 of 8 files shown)


Algue-Rythme:
As suggested in discussions:

- usage of `increase_factor` and `decrease_factor` for more explicit API
- I removed the example that compares GD algorithms (will be pushed in an other PR)
- `armijo_line_search` is now its own function that can be jitted, with loops that can be unrolled or not. Notice the code duplication in `__post_init__()` and the `_run()` of the base class: it might be worth it to encapsulate this logic in a separate method in the parent class. Significative gains in speed.
- it is in version `0.1.1` of the changelog.

Algue-Rythme:
I modified examples files in `deep_learning` accordingly to the last commit.

Algue-Rythme:
Done :)

PR #63, opened 2021-10-08
Rename step_size in stepsize

Algue-Rythme:
For consistency with `ProximalGradient` the `step_size` parameter is now named `stepsize` in `PolyakSGD`.


mblondel:
It's also used in a few examples:
```batch
$ git grep max_step_size | grep examples/
```

Algue-Rythme:
> ```batchfile
> git grep max_step_size | grep examples/
> ```

Indeed, it is now corrected.

PR #68, opened 2021-10-12
Extend Deep learning examples to other datasets.

Algue-Rythme:
I added some new datasets to Flax and Haiku examples to have more diversity (e.g Cifar).
I modified the architecture of Haiku example to make it similar to Flax.


Algue-Rythme:
Upppon suggestion I renamed "LeNet" into "MLP".

PR #100, opened 2021-11-24
FIX: examples are not running with a local GPU

fabianp:
None of the examples using tfds was able to run in my machine with
a local GPU. I think this is important, as leveraging the GPU is
important for the more large-scale datasets.

This fixes it. The same trick is used in the flax
examples.

### DIFF
--- a/examples/deep_learning/flax_image_classif.py
+++ b/examples/deep_learning/flax_image_classif.py
@@ -34,6 +34,7 @@ from jaxopt import tree_util
 import optax
 
 import tensorflow_datasets as tfds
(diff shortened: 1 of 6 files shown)


mblondel (approved):
LGTM! Thanks!

mblondel:
The [DEQ example](https://github.com/google/jaxopt/blob/main/examples/fixed_point/deep_equilibrium_model.py#L59) is also affected.

fabianp:
good point, I changed that example accordingly and also corrected some inconsistent indentation in that example

PR #104, opened 2021-12-01
IterativeRefinement to improve EqualityConstrainedQP speed

Algue-Rythme:
## What is new ?

The class `IterativeRefinement` is a wrapper to increases the precision of linear system solvers: by minimizing the residuals iteratively a high accuracy solution can be obtained from a low accuracy solver, which is useful in `float32` for ill-posed problems.

This functionnality is also available through a function in `linear_solve`, that mimic the interface of `solve_gmres,solve_bicgstab` etc...

Iterative Refinement solves `Ax=b` with the following recursion:

iterative

Where `x_0=0` and `r_0=b`. The matrix `A_bar` can be any (regularized) approximation of `A`. The linear solver on the first step can be inaccurate, as long as computing the residuals `b-Ax` is accurate.

The user can specify its own `A_bar`; otherwise we take `A_bar=A`.

## Link with EqualityConstrainedQP

This allows `EqualityConstrainedQP` to solve badly scaled problems (see new test).
The following attributes have been added to EqualityConstrainedQP:

- refine_regularization: (optional) a positive float specifying the strength of regularization.
- refine_maxiter: maximum number of refinement steps when refine_regularization is not None.

## Question

Under which section should we document the class `IterativeRefinement` ? Do we need a `Linear Solvers` section ? Or under `constrained optimization` ? Finding a feasible point (constrained problem, without objective) can be turned into an optimization problem (unconstrained minimization of residuals).

### DIFF
--- a/jaxopt/_src/eq_qp.py
+++ b/jaxopt/_src/eq_qp.py
@@ -19,17 +19,22 @@ from typing import Callable
 from typing import Optional
 from typing import Tuple
 
+from functools import partial
 from dataclasses import dataclass
 
 import jax
 import jax.numpy as jnp 
 
+from jaxopt._src import loop
 from jaxopt._src import base
 from jaxopt._src import implicit_diff as idf
-from jaxopt._src import linear_solve
-from jaxopt._src import tree_util
+from jaxopt._src.tree_util import tree_zeros_like, tree_add, tree_sub
+from jaxopt._src.tree_util import tree_add_scalar_mul, tree_scalar_mul 
+from jaxopt._src.tree_util import tree_vdot, tree_negative, tree_l2_norm
 from jaxopt._src.linear_operator import _make_linear_operator
 from jaxopt._src.cvxpy_wrapper import _check_params
+import jaxopt._src.linear_solve as linear_solve
+from jaxopt._src.iterative_refinement import IterativeRefinement
 
 
 def _make_eq_qp_optimality_fun(matvec_Q, matvec_A):
@@ -44,13 +49,13 @@ def _make_eq_qp_optimality_fun(matvec_Q, matvec_A):
   def obj_fun(primal_var, params_obj):
     params_Q, c = params_obj
     Q = matvec_Q(params_Q)
-    return (0.5 * tree_util.tree_vdot(primal_var, Q(primal_var)) +
-            tree_util.tree_vdot(primal_var, c))
+    return (0.5 * tree_vdot(primal_var, Q(primal_var)) +
+            tree_vdot(primal_var, c))
 
   def eq_fun(primal_var, params_eq):
     params_A, b = params_eq
     A = matvec_A(params_A)
-    return tree_util.tree_sub(A(primal_var), b)
+    return tree_sub(A(primal_var), b)
 
   optimality_fun_with_ineq =  idf.make_kkt_optimality_fun(obj_fun, eq_fun, ineq_fun=None)
 
@@ -69,7 +74,8 @@ class EqualityConstrainedQP(base.Solver):
   Supports implicit differentiation, matvec and pytrees.
   Can benefit from GPU/TPU acceleration.
 
-  Not as precise as CVXPY_QP.
+  Not as precise as CvxpyQP. If the algorithm diverges on some instances,
+  it might be useful to tweak the ``refine_regularization`` parameter.
 
   Attributes:
     matvec_Q: a Callable matvec_Q(params_Q, u).
@@ -77,6 +83,10 @@ class EqualityConstrainedQP(base.Solver):
     matvec_A: a Callable matvec_A(params_A, u).
       By default, matvec_A(A, u) = dot(A, u), where A = params_A.
     solve: a Callable to solve linear systems, that accepts matvecs (default: linear_solve.solve_gmres).
+    refine_regularization: a float (default: 0.) used to regularize the system.
+      Useful for badly conditioned problems that lead to divergence.
+      ``IterativeRefinement`` is used to correct the error introduced.
+    refine_maxiter: maximum number of refinement steps, when refine_regularization is not 0.
     maxiter: maximum number of iterations.
     tol: tolerance (stoping criterion).
     implicit_diff: whether to enable implicit diff or autodiff of unrolled
@@ -87,13 +97,62 @@ class EqualityConstrainedQP(base.Solver):
   matvec_Q: Optional[Callable] = None
   matvec_A: Optional[Callable] = None
   solve: Callable = linear_solve.solve_gmres
+  refine_regularization: float = 0.0
+  refine_maxiter: int = 10
   maxiter: int = 1000
   tol: float = 1e-5
   implicit_diff_solve: Optional[Callable] = None
   jit: bool = True
 
+  def _refined_solve(self, matvec, b, init, maxiter, tol):
+    # Instead of solving S x = b
+    # We solve     \bar{S} x = b
+    #
+    # S = [[  Q   A^T  
+    #         A    0 ]]
+    #
+    # \bar{S} = [[ Q + delta      A^T
+    #                A         - delta ]]
+    #
+    # Since Q is PSD, and delta a diagonal matrix,
+    # this makes \bar{S} a quasi-definite matrix.
+    #
+    # Quasi-Definite matrices are indefinite,
+    # but they are never singular (when delta > 0).
+    #
+    # This guarantees that the system is well posed in every case,
+    # even if some constraints of A are redundant.
+    #
+    # The particular form of this system is inspired by [2].
+    # Quasi-Definite matrices are a known tool in the literature on Interior Point methods [1].
+    #
+    # References:  
+    #  
+    #  [1] Vanderbei, R.J., 1995. Symmetric quasidefinite matrices.
+    #      SIAM Journal on Optimization, 5(1), pp.100-113.  
+    #  
+    #  [2] Stellato, B., Banjac, G., Goulart, P., Bemporad, A. and Boyd, S., 2020.
+    #      OSQP: An operator splitting solver for quadratic programs.
+    #      Mathematical Programming Computation, 12(4), pp.637-672.  
+
+    def matvec_qp(_, x):
+      return matvec(x)
+
+    ridge = self.refine_regularization
+
+    def matvec_regularized_qp(_, x):
+      primal, dual_eq = x
+      Stop, Sbottom = matvec(x)
+      return Stop + ridge * primal, Sbottom - ridge * dual_eq
+
+    solver = IterativeRefinement(matvec_A=matvec_qp,
+                                 matvec_A_bar=matvec_regularized_qp,
+                                 solve=partial(self.solve, maxiter=maxiter, tol=tol),
+                                 maxiter=self.refine_maxiter, tol=tol)
+    return solver.run(init_params=init, params_A=None, b=b)[0]
+
   def run(self,
-          init_params: Optional[Any] = None,
+          init_params: Optional[base.KKTSolution] = None,
           params_obj: Optional[Any] = None,
           params_eq: Optional[Any] = None) -> base.OptStep:
     """Solves 0.5 * x^T Q x + c^T x subject to Ax = b.
@@ -107,7 +166,6 @@ class EqualityConstrainedQP(base.Solver):
     Returns:
       (params, state),  where params = (primal_var, dual_var_eq, None)
     """
-    del init_params  # no warm start
     if self._check_params:
       _check_params(params_obj, params_eq)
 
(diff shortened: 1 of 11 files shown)


jaxopt/_src/iterative_refinement.py
```diff
@@ -0,0 +1,210 @@
...
+    tol: absolute tolerance for stoping criterion (default: 1e-7).
+    verbose: If verbose=1, print error at each iteration.
+    implicit_diff: whether to enable implicit diff or autodiff of unrolled iterations.
+    implicit_diff_solve: the linear system solver to use.
+    jit: whether to JIT-compile the optimization loop (default: "auto").
+    unroll: whether to unroll the optimization loop (default: "auto")
+
+  References:
+    J. H. Wilkinson. Rounding Errors in Algebraic Processes. Prentice Hall, Englewood Cliffs, NJ, 1963.
+    Moler, C.B., 1967. Iterative refinement in floating point. Journal of the ACM (JACM), 14(2), pp.316-321.
```
mblondel, line 98: Add link to https://en.wikipedia.org/wiki/Iterative_refinement

jaxopt/_src/iterative_refinement.py
```diff
@@ -0,0 +1,210 @@
...
+             params_A_bar: Optional[Any] = None):
+    if self._copy_params_A:
+      params_A_bar = params_A
+
+    A = self.matvec_A(params_A)
+    A_bar = self.matvec_A_bar(params_A_bar)
+
+    # TODO(lbethune): support preconditioners ?
+    # Could it be done by user with partial(solver, M=precond) ?
+    residual_sol = self.solve(A_bar, state.target_residuals, x0=state.init)
```
mblondel, line 135: In `cg`, the option is called `init` (https://github.com/google/jaxopt/blob/main/jaxopt/_src/linear_solve.py#L91). We should probably add an `init` option to `gmres` and other solvers instead of relying on `kwargs`...

jaxopt/_src/iterative_refinement.py
```diff
@@ -0,0 +1,210 @@
...
+                          params_A: Any,
+                          b: Any,
+                          params_A_bar: Optional[Any] = None):
+    del params_A_bar  # unused
+    return tree_l2_norm(self.optimality_fun(params, params_A, b))
+
+  def __post_init__(self):
+    # The inner solver might not support autodiff, so 
+    # we force implicit_diff in every case.
+    self.implicit_diff = True
```
mblondel, line 173: Is this necessary? (https://github.com/google/jaxopt/blob/main/jaxopt/_src/base.py#L120)

jaxopt/_src/iterative_refinement.py
```diff
@@ -0,0 +1,210 @@
...
+  Args:
+    matvec: product between ``A`` and a vector.
+    b: pytree.
+    maxiter: maximum number of refinement steps (default: 10).
+    tol: absolute tolerance residuals (default: 1e-7).
+    solve: optional solve function (default: linear_solve.solve_gmres).
+    kwargs: additional parameters for IterativeRefinement.
+
+  Returns:
+    array with same structure as ``b``.
```
mblondel, line 202:
```suggestion
    Pytree with the same structure as ``b``.
```

jaxopt/_src/iterative_refinement.py
```diff
@@ -0,0 +1,210 @@
...
+  """
+  iterative_refinement = IterativeRefinement(matvec_A=lambda _,x: matvec(x),
+                                             solve=solve,
+                                             maxiter=maxiter,
+                                             tol=tol,
+                                             **kwargs)
+  return iterative_refinement.run(init, None, b)[0]
```
mblondel, line 209: Use keyword arguments (it's not clear what None refers to).

jaxopt/_src/linear_operator.py
```diff
@@ -25,6 +25,7 @@ class DenseLinearOperator:
 
   def __init__(self, pytree):
     self.pytree = pytree
+    self.linop_type = 'dense'
```
mblondel, line 28: I'm not a fan of this.

jaxopt/_src/linear_operator.py
```diff
@@ -59,6 +60,7 @@ class FunctionalLinearOperator:
 
   def __init__(self, fun, params):
     self.fun = functools.partial(fun, params)
+    self.linop_type = 'functional'
```
mblondel, line 63: Ditto

jaxopt/_src/osqp.py
```diff
@@ -435,26 +441,34 @@ def _solve_linear_system(self, params, Q, c, A, rho_bar, state):
     x, z = params.primal
     y    = params.dual_eq  # dual variables for constraints z_bar = z;
 
-    def matvec(x_bar):
-      Qx_sigmax = tree_add_scalar_mul(Q(x_bar), self.sigma, x_bar)
-      ATAx = A.normal_matvec(x_bar)
-      return tree_add_scalar_mul(Qx_sigmax, rho_bar, ATAx)
+    if Q.linop_type == A.linop_type == 'dense':
```
mblondel, line 444: Use `isinstance(Q, DenseLinearOperator)` instead.

jaxopt/_src/tree_util.py
```diff
@@ -58,6 +58,10 @@ def tree_vdot(tree_x, tree_y):
   return tree_reduce(operator.add, vdots)
 
 
+def tree_dot(tree_x, tree_y):
```
mblondel, line 61: Add docstring (what does it mean to multiply two pytrees?)

tests/eq_qp_test.py
```diff
@@ -140,6 +143,29 @@ def matvec_A(A, tup):
...
+
+    onp.random.seed(511)
+    X = onp.random.rand(size_prob, size_prob)
+    Q = jnp.array(X @ X.T)
+    c = jnp.array(onp.random.rand(size_prob))
+
+    A = jnp.array(onp.random.rand(size_eq, size_prob))
+    b = jnp.array(onp.random.rand(size_eq))
+    
+    # The tolerance is low, but ordinary this problem cannot be solved by
```
mblondel, line 158:
```suggestion
    # The tolerance is low, but ordinarily this problem cannot be solved by
```

tests/iterative_refinement_test.py
```diff
@@ -0,0 +1,122 @@
...
+    sigma = 0.05
+    A_bar = A + sigma * noise  # perturbed system.
+    
+    x0 = onp.random.randn(n)  # expected solution.
+    b = A @ x0  # unperturbed target.
+
+    high_acc = 1e-3
+    solver = IterativeRefinement(matvec_A=None, matvec_A_bar=jnp.dot,
+                                 tol=high_acc, maxiter=100)
+    x, state = solver.run(None, A, b, A_bar)
```
mblondel, line 96: Use keyword arguments.

tests/iterative_refinement_test.py
```diff
@@ -0,0 +1,122 @@
...
+    b = onp.random.randn(n)
+
+    low_acc = 1e-1
+    high_acc = 1e-5
+
+    # Heavily regularized low acuracy solver.
+    inner_solver = partial(_linear_solve.solve_gmres, tol=low_acc, ridge=1e-3)
+
+    solver = IterativeRefinement(solve=inner_solver, tol=high_acc, maxiter=10)
+    x, state = solver.run(None, A, b)
```
mblondel, line 46: Use keyword arguments.

tests/iterative_refinement_test.py
```diff
@@ -0,0 +1,122 @@
...
+    onp.random.seed(0)
+    n = 20
+
+    A = onp.random.rand(n, n)  # invertible matrix (with high probability).
+
+    noise = onp.random.randn(n, n)
+    sigma = 0.05
+    A_bar = A + sigma * noise  # perturbed system.
+    
+    x0 = onp.random.randn(n)  # expected solution.
```
mblondel, line 90:
```suggestion
    expected = onp.random.randn(n)
```

mblondel (changes requested):
Thanks Louis!

jaxopt/_src/eq_qp.py
```diff
@@ -62,21 +67,33 @@ def optimality_fun(params, params_obj, params_eq):
   return optimality_fun
 
 
+@dataclass
+class EqualityConstrainedQPState:
```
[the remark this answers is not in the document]
Algue-Rythme: No, indeed. I am waiting for `linear_solve` solvers to return a Tuple whose second member without be infos on the algforithm (error, iter_num, etc... ). 

Not useful currently but I hope it will be in the future once linear solvers adopt the Jaxopt base.IterativeSolver interface.

jaxopt/_src/eq_qp.py
```diff
@@ -87,13 +104,32 @@ class EqualityConstrainedQP():
...
   implicit_diff_solve: Optional[Callable] = None
   jit: bool = True
 
+  def _refined_solve(self, matvec, b, init, maxiter, tol):
+    def matvec_A(_, x):
+      return matvec(x)
+
+    ridge = self.refine_regularization
+
+    def matvec_A_bar(_, x):
```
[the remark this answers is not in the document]
Algue-Rythme: I added some comments and references: it appears that quasi-definite matrices have been studied in the context of _interior point methods._ They are no stranger to quadratic programming:

Vanderbei, R.J., 1995. Symmetric quasidefinite matrices.
SIAM Journal on Optimization, 5(1), pp.100-113.

jaxopt/_src/eq_qp.py
```diff
@@ -87,13 +104,32 @@ class EqualityConstrainedQP():
...
+      p, d = x
+      p_new, d_new = matvec(x)
```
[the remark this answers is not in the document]
Algue-Rythme: You are right this variable's name is awful (and not even right): I change that !

jaxopt/_src/iterative_refinement.py
Algue-Rythme: Good call !

On the other hand the `jax.scipy.sparse.linalg` interface (that the user might be more familar with) uses `x0`, and that's what is used under the hood for `jaxopt.linear_solve.solve_gmres`.

There is a mismatch between those two interfaces.

jaxopt/_src/iterative_refinement.py
Algue-Rythme: It seems that the `run` method is wrapped only if `implicit_diff=True` but I might be wrong: https://github.com/google/jaxopt/blob/648f2f2553b1d91eaf54577ea0413ad64729089f/jaxopt/_src/base.py#L172

jaxopt/_src/tree_util.py
Algue-Rythme: I added:

_Compute leaves-wise dot product between pytree of arrays._
  
  _Useful to store block diagonal linear operator: each leaf of the tree
  corresponds to a block._

jaxopt/_src/iterative_refinement.py
mblondel: `getattr(self, "implicit_diff", True)` returns `True` if "implicit_diff" is not present.

Algue-Rythme: Oh yeah I didn't realized that !

Algue-Rythme:
Fixup commit; ready for merge after final review.

jaxopt/_src/eq_qp.py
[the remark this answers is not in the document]
mblondel: Let's delete it for the time being then.

jaxopt/_src/iterative_refinement.py
```diff
@@ -0,0 +1,224 @@
...
+  """
+  iter_num: int
+  error: float
+  target_residuals: Any
+  init: Optional[Any]
+  # TODO(lbethune): in the future return the state of the internal
+  # solver (iter_num, error) as part of the current state.
+
+
+@dataclass
```
mblondel, line 60:
```suggestion
@dataclass(eq=False)
```

jaxopt/_src/iterative_refinement.py
```diff
@@ -0,0 +1,224 @@
...
+  jit: base.AutoOrBoolean = "auto"
+  unroll: base.AutoOrBoolean = "auto"
+
+  def init_state(self,
+                 init_params: Any,
+                 params_A: Any,
+                 b: Any,
+                 params_A_bar: Any = None):
+    return IterativeRefinementState(
+      iter_num=0,
```
mblondel, line 122:
```suggestion
      iter_num=jnp.asarray(0),
```

jaxopt/_src/iterative_refinement.py
```diff
@@ -0,0 +1,224 @@
...
+      error=jnp.inf,
```
mblondel, line 123:
```suggestion
      error=jnp.asarray(jnp.inf),
```

mblondel (changes requested):
Some more comments. Let's merge after this. Also, could you squash the commits?

Algue-Rythme:
Commits squashed + modifications made as suggested.

PR #106, opened 2021-12-01
Make solver.update jittable and ensure output states are consistent.

mblondel:


jaxopt/_src/test_util.py
```diff
@@ -155,3 +155,22 @@ def lsq_linear_cube_osp(X, y, l, tol=1e-10, max_iter=None):
...
+    raise ValueError("state1 and state2 should have the same number of "
+                     "attributes.")
+
+  for attr1, attr2 in zip(state1._fields, state2._fields):
+    if attr1 != attr2:
+      raise ValueError("Attribute names do not agree: %s and %s." % (attr1,
+                                                                     attr2))
+
+    value1_cls_name = getattr(state1, attr1).__class__.__name__
+    value2_cls_name = getattr(state2, attr2).__class__.__name__
```
fllinares, line 171: Naive question: why not use `type` here?

mblondel: No big reason. Following your comment, I switched to type.

jaxopt/_src/anderson.py
```diff
@@ -173,8 +173,8 @@ def init_state(self,
     residuals_history = tree_map(jnp.zeros_like, params_history)
     residual_gram = jnp.zeros((m,m))
 
-    return AndersonState(iter_num=0,
-                         error=jnp.inf,
+    return AndersonState(iter_num=jnp.asarray(0),
+                         error=jnp.asarray(jnp.inf),
```
froystig, line 177: It seems worthwhile to document—in a comment or even in this PR's description—the reason for sending these scalars through `jnp.asarray`. My understanding is that we would like to return a state struct that is the same whether or not the update took place on device (via `jit`). Is that correct?

In some situations, we might avoid writing expressions like `jnp.asarray(1.)` whenever possible. If we're not under a `jit`, the expression allocates a scalar on the default device, which could be an accelerator. Considering this, what is this degree of consistency buying us, and is it worth this cost?

mblondel: The goal was to avoid this:
```
jitted_update = jax.jit(solver.update)
state = solver.init_state(params)
# here state contains floats
params, state = jitted_update(params, state)
# here state contains arrays due to jit
params, state = jitted_update(params, state)  # recompilation occurs
```
I agree it could be nice to add a comment but not sure if we should repeat the same comment in all solvers.

froystig: I believe that if we use `onp.asarray(...)` instead of `jnp.asarray`, we will on the one hand keep these scalars in host memory, and on the other hand be consistent for `jit` (in the sense of shape/dtype), such that it won't recompile.

mblondel: @fllinares and I would like to understand this better. Since this PR is already an improvement, let's merge and explore `onp.asarray` separately (potentially we could make measurements).

fllinares (approved):
On my side, everything LGTM! Thanks a lot Mathieu!

jaxopt/_src/anderson.py
froystig: I agree. Sounds good!

froystig (approved):

PR #113, opened 2021-12-04
Prevent recompilation of loops in `solver.run` if executing without jit.

fllinares:
#### Context

An issue was found whereby executing the `run` method of `IterativeSolver`s without `jit` would cause `lax.while_loop` or `lax.scan` to be retraced in each subsequence call. This was due to `cond_fun` and `body_fun` being (re)defined inside the private method `_run` of `IterativeSolver`.

#### Solution and caveats

The functions `cond_fun` and `body_fun` have been defined as methods of `IterativeSolver`. This solved the retracing issues. However, it causes two important changes in the way subclasses of `IterativeSolver` behave, namely
1. Methods of `IterativeSolver` must support serving as inputs to JAX transforms directly (see point below).
2. Positional and keyword arguments of `run` are no longer passed to `body_fun` with a closure but rather as explicit inputs. Hence, they must support JAX tracing.

#### JAX-transforming methods of `IterativeSolver` / bypassing a Python 3.7 dataclass issue

We found that applying JAX transformations such as `jax.jit` directly to methods of `IterativeSolver` raises a `TypeError: unhashable type` exception in Python 3.7 whenever the solver is a Python dataclass with `eq=True`. This problem does *not* occur in Python 3.8 nor Python 3.9. The issue is particularly problematic for this PR, as the same limitation occurs when passing the newly-defined-as-methods `body_fun` and `cond_fun` to `lax.while_loop` or `lax.scan`.

*To solve this issue, we temporarily set `eq=False` in all JAXopt solvers. This might change in the future if either an alternative solution is found, or support for Python 3.7 is eventually dropped.*

#### Other

This PR is joint work with @mblondel and @Algue-Rythme.

### DIFF
--- a/jaxopt/_src/base.py
+++ b/jaxopt/_src/base.py
@@ -59,6 +59,13 @@ class Solver(abc.ABC):
 
   In addition, we will assume that the solver defines a notion of optimality:
     - `pytree = optimality_fun(params, *args, **kwargs)
+
+  Note: subclasses of `Solver` which are also Python dataclasses must disable
+  the generation of an `__eq__` method (`eq=False`) for compatibility with
+  Python 3.7. Not doing so results in the generation of a `__hash__` method
+  that clashes with JAX transformations when applied directly to methods of
+  the class. This restriction will be lifted in the future when support for
+  Python 3.7 is dropped and / or another solution is found.
   """
 
   @abc.abstractmethod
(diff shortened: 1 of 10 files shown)


mblondel:
Thanks a lot @fllinares! LGTM! I wonder if we can come up with a non-regression test that failed before your changes.

BTW, there seems to be some failures in Anderson acceleration, not sure what's going on.

mblondel (approved):

PR #114, opened 2021-12-06
float32/float64 tests for OSQP

Algue-Rythme:
Add tests in `float32` for OSQP; create files `osqp_float32_test.py` and `osqp_float64_test.py`.


Algue-Rythme:
Tests are passing both in `float32` and `float64` without tuning of `tolerance` parameter: this PR is useless there is no need for code duplication.

PR #119, opened 2021-12-07
ENH: make flax_resnet.py faster and more accurate across the board

fabianp:
Most changes are geared towards ensuring we give strong and
efficient baselines by default, without parameter tuning. The most
dramatic gains are in the CIFAR* datasets.

Currently, the train/test accuracy for the different datasets with
default flags are:

  - MNIST: 0.99/0.99 (this one didn't change much)
  - FASHION_MNIST: 0.94/0.88 with a Time elapsed of 0:03:09
       (vs 0.71/0.69 for previous defaults).
  - E_MNIST: 0.85/0.85 with a Time elapsed of 0:25:56
       (vs 0.31/0.37 for previous defaults).
  - CIFAR10: 0.87/0.75 with Time elapsed of 0:02:36
       (vs 0.23/0.22 for previous default architecture and
        0.29/0.32 for resnet18).
  - CIFAR100: 0.72/0.39, with a time elapsed: 0:02:31
        (vs 0.07/0.06 for previous default architecture and
        0.11/0.09 for resnet18).

Here, time elapsed is on a workstation with a GeForce GTX 1080 GPU. It
will certainly be slower on CPU, although I expect most people aiming to
train large resnets to have access to a GPU.

More precise, main changes are:

  * Refactored the printing statements to be able to jit compile the
    update rule. It now runs very efficiently on GPU.

  * Replace maxiter by epochs. The maxiter of 100 is nowhere near
    to give reasonable accuracy on CIFAR10, and I believe that setting
    the limit in terms of epochs is both more common and easier to set
    when considering several datasets in the same example.

  * Use resnet18 as default. resnet1 gives 0.4 test accuracy on cifar10
    after 30 epochs, while resnet18 goes up to 0.7.

  * Print running time. It's been useful for me to compare the efficiency
    of different approaches.

### DIFF
--- a/examples/deep_learning/flax_resnet.py
+++ b/examples/deep_learning/flax_resnet.py
@@ -19,6 +19,7 @@ Resnet example with Flax and JAXopt.
 
 from absl import app
 from absl import flags
+from datetime import datetime
 
 from functools import partial
 from typing import Any, Callable, Sequence, Tuple
@@ -44,11 +45,11 @@ dataset_names = [
 
 
 flags.DEFINE_float("l2reg", 1e-4, "L2 regularization.")
-flags.DEFINE_float("learning_rate", 0.1, "Learning rate.")
-flags.DEFINE_integer("maxiter", 100, "Maximum number of iterations.")
+flags.DEFINE_float("learning_rate", 0.2, "Learning rate.")
+flags.DEFINE_integer("epochs", 10, "Number of passes over the dataset.")
 flags.DEFINE_float("momentum", 0.9, "Momentum strength.")
 flags.DEFINE_enum("dataset", "mnist", dataset_names, "Dataset to train on.")
-flags.DEFINE_enum("model", "resnet1", ["resnet1", "resnet18", "resnet34"],
+flags.DEFINE_enum("model", "resnet18", ["resnet1", "resnet18", "resnet34"],
                   "Model architecture.")
 flags.DEFINE_integer("train_batch_size", 256, "Batch size at train time.")
 flags.DEFINE_integer("test_batch_size", 1024, "Batch size at test time.")
@@ -150,6 +151,7 @@ def main(argv):
                             batch_size=FLAGS.test_batch_size)
   input_shape = (1,) + ds_info.features["image"].shape
   num_classes = ds_info.features["label"].num_classes
+  iter_per_epoch = ds_info.splits['train'].num_examples // FLAGS.train_batch_size
 
   # Set up model.
   if FLAGS.model == "resnet1":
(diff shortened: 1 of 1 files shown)


mblondel (approved):
LGTM, thanks a lot Fabian!

fabianp:
@mblondel : do you think I should switch the maxiter -> epochs in the related examples for consistency? I'm thinking flax_image_classif.py and perhaps the haiku* ones

mblondel:
@fabianp That could be nice. That said, I would like to keep an example that uses `run_iterator`. Maybe we should try `id_print` within `pre_update` in these examples.

examples/deep_learning/flax_resnet.py
```diff
@@ -192,48 +194,51 @@ def loss_fun(params, l2reg, data, aux):
...
+      test_loss = jax.device_get(test_loss)
+      # time elapsed without microseconds
+      time_elapsed = (datetime.now().replace(microsecond=0) - start)
+
+      print(f"[Epoch {state.iter_num // (iter_per_epoch+1)}/{FLAGS.epochs}] "
+            f"Train acc: {train_acc:.3f}, train loss: {train_loss:.3f}. "
+            f"Test acc: {test_acc:.3f}, test loss: {test_loss:.3f}. "
+            f"Time elapsed: {time_elapsed}")
+
+    params, state = jax.jit(solver.update)(params=params,
```
mblondel, line 235: I hadn't noticed but you're recompiling on every iteration. Also, when/if we implement #117, this will no longer be necessary. CC @fllinares

fabianp: huh, good catch. Performance wise I didn't observe much overhead, with the GPU utilization around 95% . 

Is there a workaround I should use? Perhaps define an ```_update = jax.jit(solver.update)``` function outside of the for loop?

fabianp:
> @fabianp That could be nice. That said, I would like to keep an example that uses `run_iterator`. Maybe we should try [id_print](https://jax.readthedocs.io/en/latest/jax.experimental.host_callback.html#jax.experimental.host_callback.id_print) within `pre_update` in these examples.

Should be doable since I'm just computing the max_iter necessary to reach X epochs, not changing the structure of the for loop

examples/deep_learning/flax_resnet.py
mblondel: Yep exactly (maybe use the name `jitted_update` to be more explicit)

fabianp:
I added the jitted_update to avoid recompilation. i'll open an issue for the renaming of maxiter-> epochs in other examples

PR #121, opened 2021-12-09
Add OSQP to SVMs examples.

Algue-Rythme:
The two examples of SVM `binary_kernel_svm_with_intercept` and `multiclass_linear_svm` are now solved with ``BoxOSQP`` and ``OSQP`` on toy datasets.

The `binary_kernel_svm_with_intercept` example also exhibit a `LinearKernel` class that illustrates a possible API for kernels. See also the (unmerged) branch: https://github.com/Algue-Rythme/jaxopt/blob/svmfast/jaxopt/_src/kernels.py

### DIFF
--- a/examples/constrained/binary_kernel_svm_with_intercept.py
+++ b/examples/constrained/binary_kernel_svm_with_intercept.py
@@ -19,52 +19,63 @@ Binary kernel SVM with intercept.
 The dual objective of binary kernel SVMs with an intercept contains both
 box constraints and an equality constraint, making it challenging to solve.
 The state-of-the-art algorithm to solve this objective is SMO (Sequential
-minimal optimization). We nevertheless demonstrate in this example how to solve
+minimal optimization). 
+
+We nevertheless demonstrate in this example how to solve
 it by projected gradient descent, by projecting on the constraint set
-using projection_box_section.
+using projection_box_section.  
+  
+Since the dual objective is a Quadratic Program we show how to solve
+it with BoxOSQP too.
 """
 
 from absl import app
+from absl import flags
+
+import jax
 import jax.numpy as jnp
 from jaxopt import projection
 from jaxopt import ProjectedGradient
+from jaxopt import BoxOSQP
+
 import numpy as onp
 from sklearn import datasets
 from sklearn import preprocessing
 from sklearn import svm
 
 
-def objective_fun(beta, lam, K, y):
-  """Dual objective of binary kernel SVMs with intercept."""
-  # The dual objective is:
-  # fun(beta) = 0.5 beta^T K beta - beta^T y
-  # subject to
-  # sum(beta) = 0
-  # 0 <= beta_i <= C if y_i = 1
-  # -C <= beta_i <= 0 if y_i = -1
-  # where C = 1.0 / lam
-  return 0.5 * jnp.dot(beta, jnp.dot(K, beta)) - jnp.dot(beta, y)
+flags.DEFINE_float("lam", 0.5, "Regularization parameter. Must be positive.")
+flags.DEFINE_float("tol", 1e-6, "Tolerance of solvers.")
+flags.DEFINE_integer("num_samples", 30, "Size of train set.")
+flags.DEFINE_integer("num_features", 5, "Features dimension.")
+flags.DEFINE_bool("verbose", False, "Verbosity.")
+FLAGS = flags.FLAGS
 
 
-def binary_kernel_svm_skl(K, y, lam, tol=1e-5):
-  svc = svm.SVC(kernel="precomputed", C=1.0 / lam).fit(K, y)
+def binary_kernel_svm_skl(X, y, C):
+  print(f"Solve SVM with sklearn.svm.SVC: ")
+  K = jnp.dot(X, X.T)
+  svc = svm.SVC(kernel="precomputed", C=C, tol=FLAGS.tol).fit(K, y)
   dual_coef = onp.zeros(K.shape[0])
   dual_coef[svc.support_] = svc.dual_coef_[0]
   return dual_coef
 
 
-def main(argv):
-  del argv
+def binary_kernel_svm_pg(X, y, C):
+  print(f"Solve SVM with Projected Gradient: ")
 
-  # Prepare data.
-  X, y = datasets.make_classification(n_samples=20, n_features=5,
-                                      n_informative=3, n_classes=2,
-                                      random_state=0)
-  X = preprocessing.Normalizer().fit_transform(X)
-  y = y * 2 - 1  # Transform labels from {0, 1} to {-1, 1}.
-  lam = 1.0
-  C = 1./ lam
-  K = jnp.dot(X, X.T)  # Use a linear kernel.
+  def objective_fun(beta, X, y):
+    """Dual objective of binary kernel SVMs with intercept."""
+    # The dual objective is:
+    # fun(beta) = 0.5 beta^T K beta - beta^T y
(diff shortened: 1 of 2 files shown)


examples/constrained/binary_kernel_svm_with_intercept.py
```diff
@@ -19,72 +19,193 @@
...
-  """Dual objective of binary kernel SVMs with intercept."""
+flags.DEFINE_float("lam", 0.5, "Regularization parameter. Must be positive.")
+flags.DEFINE_float("tol", 1e-6, "Tolerance of solvers.")
+flags.DEFINE_integer("num_samples", 30, "Size of train set.")
+flags.DEFINE_integer("num_features", 5, "Features dimension.")
+flags.DEFINE_bool("verbose", False, "Verbosity.")
+FLAGS = flags.FLAGS
+
+
+class LinearKernel:
```
mblondel, line 55: I would prefer if we don't introduce classes for this simple case. This obfuscates the code IMO...

examples/constrained/binary_kernel_svm_with_intercept.py
```diff
@@ -19,72 +19,193 @@
...
+
+  def get_dense(self, params):
+    X, Y = params
+    return X @ Y.T
+
+
+def binary_kernel_svm_skl(X, y, C, kernel):
+  print(f"Solve SVM with sklearn.svm.SVC: ")
+  kernel_params = kernel.init_params(X)
+  K = kernel.get_dense(kernel_params)
```
mblondel, line 96: K = jnp.dot(X, X.T)

examples/constrained/binary_kernel_svm_with_intercept.py
```diff
@@ -19,72 +19,193 @@
...
+    """Dual objective of binary kernel SVMs with intercept."""
+    # The dual objective is:
+    # fun(beta) = 0.5 beta^T K beta - beta^T y
+    # subject to
+    # sum(beta) = 0
+    # 0 <= beta_i <= C if y_i = 1
+    # -C <= beta_i <= 0 if y_i = -1
+    # where C = 1.0 / lam
+    # and K = X X^T
+    Kbeta = kernel.matvec(kernel_params, beta)
```
mblondel, line 116: Kbeta = jnp.dot(X, jnp.dot(X.T beta))

examples/constrained/binary_kernel_svm_with_intercept.py
```diff
@@ -19,72 +19,193 @@
...
+  svc = svm.SVC(kernel="precomputed", C=C, tol=FLAGS.tol).fit(K, y)
+  dual_coef = onp.zeros(K.shape[0])
+  dual_coef[svc.support_] = svc.dual_coef_[0]
+  return dual_coef
+
+
+def binary_kernel_svm_pg(X, y, C, kernel):
+  print(f"Solve SVM with Projected Gradient: ")
+
+  def objective_fun(beta, kernel_params, y):
```
mblondel, line 106: I don't think we need `kernel_params` for the simple case of the linear kernel.

examples/constrained/binary_kernel_svm_with_intercept.py
```diff
@@ -19,72 +19,193 @@
...
 
-def binary_kernel_svm_skl(K, y, lam, tol=1e-5):
-  svc = svm.SVC(kernel="precomputed", C=1.0 / lam).fit(K, y)
-  dual_coef = onp.zeros(K.shape[0])
-  dual_coef[svc.support_] = svc.dual_coef_[0]
-  return dual_coef
+  kernel_params = kernel.init_params(X)
+
+  def matvec_Q(kernel_params, beta):
+    return kernel.matvec(kernel_params, beta)
```
mblondel, line 154: return jnp.dot(X, jnp.dot(X.T, beta))

examples/constrained/binary_kernel_svm_with_intercept.py
```diff
@@ -19,72 +19,193 @@
...
+  
+  def matvec_A(_, beta):
+    return beta, jnp.sum(beta)
```
mblondel, line 157: Why do you return a tuple here?

examples/constrained/binary_kernel_svm_with_intercept.py
```diff
@@ -19,72 +19,193 @@
...
+  
+  l = -jax.nn.relu(-y * C), 0.
```
mblondel, line 159: Same question. I dont' understand why it's a tuple here.

examples/constrained/multiclass_linear_svm.py
```diff
@@ -15,59 +15,130 @@
 """
 Multiclass linear SVM (without intercept).
 ==========================================
+
+This quadratic program can be solved either with OSQP either with Block coordinate descent.
```
mblondel, line 19:
```suggestion
This quadratic program can be solved either with OSQP or with block coordinate descent.
```

mblondel (changes requested):
Thanks Louis. My main comment is that I would remove the linear kernel class for now. We can revisit these ideas when we actually introduce a kernel API based on unified linear operators.

examples/constrained/binary_kernel_svm_with_intercept.py
Algue-Rythme: They are two types of constraints:

```python
0 <= y*beta <= C
sum_i beta_i = 0
```

So it feels natural to return a tuple: the first element corresponds to identity matrix (in matvec form) and the other to the vector full of ones (in mnatvec form once again).

The shapes of the output of `matvec_A` and of `l,u` must agree. Since the algorithm supports pytrees (which include tuples) transparently I figured it was maybe more clear to use them. We could have returned vector by concatenating data but it's uglier IMO.

examples/constrained/binary_kernel_svm_with_intercept.py
Algue-Rythme: I agree; let's remove them for now then.

examples/constrained/binary_kernel_svm_with_intercept.py
mblondel: Ok thanks for the clarification. It's just a matter of writing a comment explaining how this works then.

examples/constrained/binary_kernel_svm_with_intercept.py
```diff
@@ -77,14 +88,88 @@ def proj(beta, C):
...
+  # 0 <= beta_i <= C if y_i = 1
+  # -C <= beta_i <= 0 if y_i = -1
+  # where C = 1.0 / lam
+
+  print(f"Solve SVM with OSQP: ")
+
+  def matvec_Q(X, beta):
+    return jnp.dot(X, jnp.dot(X.T,  beta))
+  
+  # There is two types of constraints:
```
mblondel, line 113:
```suggestion
  # There are two types of constraints:
```

examples/constrained/binary_kernel_svm_with_intercept.py
```diff
@@ -77,14 +88,88 @@ def proj(beta, C):
...
+  #   0 <= y_i * beta_i <= C     (1)
+  # and:
+  #   sum(beta) = 0              (2)
+  # The first one involves identity matrix over the betas.
```
mblondel, line 117:
```suggestion
  # The first one involves the identity matrix over the betas.
```

mblondel (approved):
Thanks Louis! LGTM apart from the minor remarks below. Let's merge after you address them!

PR #122, opened 2021-12-09
add citing information to the documentation

fabianp:
I just realized that the citing information is only in the github README but not in the online documentation. It seems important to me to add it to the documentation, since non-dev users might never read the github README page


mblondel (approved):
LGTM, good catch and thanks Fabian!

PR #126, opened 2021-12-10
Add LU factorization in OSQP + refactorize CG.

Algue-Rythme:
### What's new ?

LU factorization is now available to solve the equality constrained linear system that arises in OSQP.  
The user choses the solver with the `eq_qp_solve` option that accepts a **string**, either 'cg' (for _conjugate gradient_), 'cg+jacobi' (conjugate gradient with _jacobi preconditoner_) or 'lu' (_lu factorization_).  
  
### What changes ?

The following parameters in the constructor of BoxOSQP (and OSQP) have been removed:
- eq_qp_maxiter
- eq_qp_tol
- eq_qp_preconditioner

They were specific to conjugate gradient and do not make sense to be part of BoxOSQP class. Beside that, default values (suggested in OSQP's paper) were always successful on all problems, since authors of OSQP proposed several heuristics.  
  
### What is lost ?

The user can no longer provide its own linear solver easily. It was too much of a burden to maintain in an API consistent with conjugate gradient and lu factorization (see discussion below on design choices). Moreover, previous tests (no longer part of `main` branch) showed that `gmres` and `bicgstab` were less stable and slower than conjugate gradient. This is also suggested by the choices made by the authors of OSQP.  
  
If the user want to add support for another solver it must conform to `OSQPLinearSolver` (private and undocumented) API. I believe the default solutions proposed (LU or CG) are sufficient for now (it covers CPU/GPU implementations of OSQP) and it is not worth complexifying the code further by supporting too much options. 
  
The `Preconditioner` API have been removed: no other preconditoner than Jacobi were found in my (superficial) literature review that would have been relevant for our use case.

### Design choices

Contrary to what I wrongly state in oral discussions with @mblondel, the LU factorization solver cannot be reformulated as a preconditioner. Moreover it does not support warm start, nor preconditioning. Hence the BoxOSQPState contained attributes `eq_qp_sol` and `params_precond` that are irrelevant in the case of LU factorization. However they need `lu_factor` attributes to be carried during computations.  
  
Consequently:
- the linear system of OSQP has specific form, that _we need to use_ in order to get significant speed improvements over general purpose solvers (i.e the matrix is PSD, Jacobi preconditioning has a "nice" form easy to compute, etc ... ).
- CG needs to carry 1) previous solution (for warm start) and 2) jacobi preconditioner (optional)
- LU facotrization needs to carry LU factors

This motivates **introducing a new attribute** `solver_state` in `BoxOSQPState` that aims to carry those informations (and removing `eq_qp_sol` and `params_precond`). They are updated when the system itself is updated, which **only happens when the stepsize is modified**.  
  
Hence I introduced the abstract class `OSQPLinearSolver` for the refactorization of the code. It exhibits three methods: `init_state` (i.e LU factorization or build Jacobi preconditioner), and `update_stepsize` method (i.e update LU factors, or preconditioner) and finally the `solve_eq_qp` method, that will either perform LU solve or conjugate gradient with warm start.  
  
Its two children classes are `OSQPIndirectSolver` (which default to CG but can indeed support other iterative solvers of Jax) and `OSQPLUSolver` (that supports LU factorization but could be easily adapted to other direct methods based on matrix factorization).  

### Test

LU factorization seems efficient on matrix with less than 1e8 entries. In the future we could support bigger matrices by using the experimental `jax.sparsify` module and sparse-friendly factorization routines (a big TODO indeed, but that would fit naturally in the current state of the API without invasive modifications).
  
### Conclusion

This refactorization makes the code of BoxOSQP in overall more clear by separating the OSQP algorithm itself (ADMM on steroids) and the linear solver, but by constraining linear solvers to a specific form. This is the price to pay for performance (a "no free lunch theorem" for maintainability). The user loses the ability to manually tune the parameters of those linear solvers, but from my previous experiences it is usually an unnecessary feature for OSQP which provides already good heuristics.


jaxopt/_src/osqp.py
```diff
@@ -93,78 +93,158 @@ class BoxOSQPState(NamedTuple):
...
   @abstractmethod
-  def update_stepsize(self, precond_params, rho_bar):
-    """Update preconditioner parameters when rho_bar changes; should be efficient."""
+  def run(self, b, osqp_state):
     pass
 
 
-class NoPreconditioner(BoxOSQPPreconditioner):
-  """Dummy class for no-preconditioning."""
+@dataclass(eq=True)
```
mblondel, line 133: Shouldn't this be `eq=False`?

jaxopt/_src/osqp.py
```diff
@@ -93,78 +93,158 @@ class BoxOSQPState(NamedTuple):
...
+class OSQPIndirectSolver:
+  """Indirect solver for Equality Constrained linear system in OSQP.
+
+  Use indirect solver with warm start, with optional support for Jacobi preconditioning.
+  Default to conjugate gradient descent.
```
mblondel, line 138:
```suggestion
  Defaults to conjugate gradient descent.
```

jaxopt/_src/osqp.py
```diff
@@ -93,78 +93,158 @@ class BoxOSQPState(NamedTuple):
...
     pass
 
 
-class NoPreconditioner(BoxOSQPPreconditioner):
-  """Dummy class for no-preconditioning."""
+@dataclass(eq=True)
+class OSQPIndirectSolver:
+  """Indirect solver for Equality Constrained linear system in OSQP.
+
+  Use indirect solver with warm start, with optional support for Jacobi preconditioning.
```
mblondel, line 137:
```suggestion
  Uses an indirect solver with warm start, with optional support for Jacobi preconditioning.
```

jaxopt/_src/osqp.py
```diff
@@ -93,78 +93,158 @@ class BoxOSQPState(NamedTuple):
...
+  Default to conjugate gradient descent.
+
+  Attributes:
+    solve: linear system solver to use (default: jax.scipy.sparse.linalg.cg).
```
mblondel, line 141: Use JAXopt's functions instead.

jaxopt/_src/osqp.py
```diff
@@ -93,78 +93,158 @@ class BoxOSQPState(NamedTuple):
...
+                     x0=prev_sol,
+                     M=lambda x: self._matvec_precond(solver_state, x),
+                     maxiter=self.maxiter,
+                     atol=atol,
+                     tol=self.tol)[0]
+    solver_state = (sol, (params_Q, params_A, sigma, rho_bar), state_precond)
+    return sol, solver_state
+
+
+@dataclass(eq=True)
```
mblondel, line 214: Same

jaxopt/_src/osqp.py
```diff
@@ -93,78 +93,158 @@ class BoxOSQPState(NamedTuple):
...
+class OSQPLUSolver:
+  """Solver based on LU factorization in OSQP.
+
+  Update LU factors when stepsize rho_bar changes."""
```
mblondel, line 218:
```suggestion
  Updates LU factors when stepsize rho_bar changes."""
```

jaxopt/_src/osqp.py
```diff
@@ -758,34 +812,34 @@ class OSQPState(NamedTuple):
...
-  Meant as drop-in replacement for CvxpyQP.  
-  No support for matvec and pytrees. Supports jit and vmap.  
-  
-  CvxpyQP is more precised and should be preferred on CPU.  
-  OSQP can be quicker than CvxpyQP when GPU/TPU are available.  
-  
+  Meant as drop-in replacement for CvxpyQP.
+  No support for matvec and pytrees. Supports jit and vmap.
+
+  CvxpyQP is more precised and should be preferred on CPU.
```
mblondel, line 818:
```suggestion
  CvxpyQP is more precise and should be preferred on CPU.
```

mblondel (changes requested):

PR #128, opened 2021-12-12
L-BFGS implementation

mblondel:
- [x] basic algorithm
- [x] reusable line search module
- [x] support for pytrees
- [x] tricks of the trade

### DIFF
--- a/docs/line_search.rst
+++ b/docs/line_search.rst
@@ -1,0 +1,79 @@
+Line search
+===========
+
+Given current parameters :math:`x_k` and a descent direction :math:`p_k`,
+the goal of a line search method is to find a step size :math:`\alpha_k`
+such that the one-dimensional function
+
+.. math::
+
+  \varphi(\alpha_k) \triangleq f(x_k + \alpha_k p_k)
+
+is minimized or at least a sufficient decrease is guaranteed.
+
+Sufficient decrease and curvature conditions
+--------------------------------------------
+
+Exactly minimizing :math:`\varphi` is often computationally costly.
+Instead, it is often preferred to search for :math:`\alpha_k` satisfying certain conditions.
+One example of these conditions are the **Wolfe conditions**
+
+.. math::
+
+    \begin{aligned}
+    f(x_k + \alpha_k p_k) &\le f(x_k) + c_1 \alpha_k \nabla f(x_k)^\top p_k \\
+    \nabla f(x_k + \alpha_k p_k)^\top p_k &\ge c_2 \nabla f(x_k)^\top p_k
+    \end{aligned}
+
+where :math:`0 < c_1 < c_2 < 1`. These conditions are explained in greater detail in
+Nocedal and Wright, see equations (3.6a) and (3.6b) there.
+
+A step size may satisfy the Wolfe conditions without being particularly close
+to a minimizer of :math:`\varphi` (Nocedal and Wright, Figure 3.5).  The
+curvature condition in the second equation can be modified to force the step
+size to lie in at least a broad neighborhood of a stationary point of
+:math:`\varphi`. Combined with the sufficient decrease condition in the first
+equation, these are known as the **strong Wolfe conditions**
+
+.. math::
+
+    \begin{aligned}
+    f(x_k + \alpha_k p_k) &\le f(x_k) + c_1 \alpha_k \nabla f(x_k)^\top p_k \\
+    |\nabla f(x_k + \alpha_k p_k)^\top p_k| &\le c_2 |\nabla f(x_k)^\top p_k|
+    \end{aligned}
+
+where again :math:`0 < c_1 < c_2 < 1`. See Nocedal and Wright, equations (3.7a) and (3.7b).
+
+Algorithms
+----------
+
+.. autosummary::
+  :toctree: _autosummary
+
+    jaxopt.BacktrackingLineSearch
+
+The :class:`BacktrackingLineSearch <jaxopt.BacktrackingLineSearch>` algorithm
+iteratively reduces the step size by some decrease factor until the conditions
+above are satisfied. Example::
+
+    ls = BacktrackingLineSearch(fun=fun, maxiter=20, condition="strong-wolfe",
+                                decrease_factor=0.8)
+    stepsize, state = ls.run(init_stepsize=1.0, params=params,
+                             descent_direction=descent_direction,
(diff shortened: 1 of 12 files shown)


jaxopt/_src/lbfgs.py
```diff
@@ -0,0 +1,286 @@
...
+  r, beta = jax.lax.scan(ihp_body_left,
+                         r,
+                         (s_history, y_history, rho_history, alpha))
+
+  return r
+
+
+def update_history(history, new_value):
+  """Shift past elements to the left and add new elements at the end."""
+  return jnp.roll(history, -1, axis=0).at[-1].set(new_value)
```
shoyer, line 100: In principle, this means each update requires a full copy of the stored history vectors. My guess is that it could be significantly more efficient to treat history as a rolling buffer, and only setting a single vector. Maybe worth putting down as a "TODO".

mblondel: I left a todo for now. How would you implement this in practice?

mblondel:
This PR is now ready for review. I finished implementing L-BFGS with native support for pytrees. Currently, it uses a simple backtracking line search. I'll add the [zoom line search](https://github.com/google/jax/blob/main/jax/_src/scipy/optimize/line_search.py) from JAX core in another PR.

docs/line_search.rst
```diff
@@ -0,0 +1,78 @@
...
+  \varphi(\alpha_k) \triangleq f(x_k + \alpha_k p_k)
+
+is minimized or at least a sufficient decrease is guaranteed.
+
+Sufficient decrease and curvature conditions
+--------------------------------------------
+
+Exactly minimizing :math:`\varphi` is often computationally costly.
+Instead, it is often preferred to search for :math:`\alpha_k` satisfying certain conditions.
+The **Wolfe conditions** are
```
fabianp, line 19:
```suggestion
One example of these conditions are the Wolfe conditions. The **Wolfe conditions** are
```

docs/line_search.rst
```diff
@@ -0,0 +1,78 @@
...
+
+.. math::
+
+    \begin{aligned}
+    f(x_k + \alpha_k p_k) &\le f(x_k) + c_1 \alpha_k \nabla f(x_k)^\top p_k \\
+    \nabla f(x_k + \alpha_k p_k)^\top p_k &\ge c_2 \nabla f(x_k)^\top p_k
+    \end{aligned}
+
+where :math:`0 < c_1 < c_2 < 1`. See Nocedal and Wright, equations (3.6a) and (3.6b).
```
fabianp, line 28:
```suggestion
where :math:`0 < c_1 < c_2 < 1`. These conditions are explained in greater detail in Nocedal and Wright, equations (3.6a) and (3.6b).
```

docs/unconstrained.rst
```diff
@@ -45,8 +46,10 @@ Instantiating and running the solver
 Continuing the ridge regression example above, gradient descent can be
 instantiated and run as follows::
 
-  gd = jaxopt.GradientDescent(fun=ridge_reg_objective, maxiter=500)
-  res = gd.run(init_params, l2reg=l2reg, X=X, y=y)
+  # solver = jaxopt.GradientDescent(fun=ridge_reg_objective, maxiter=500)
```
fabianp, line 49: is the commenting intentional? seems strange in the doc

jaxopt/_src/lbfgs.py
```diff
@@ -0,0 +1,312 @@
...
+
+    implicit_diff: whether to enable implicit diff or autodiff of unrolled
+      iterations.
+    implicit_diff_solve: the linear system solver to use.
+
+    jit: whether to JIT-compile the optimization loop (default: "auto").
+    unroll: whether to unroll the optimization loop (default: "auto").
+
+    verbose: whether to print error on every iteration or not.
+      Warning: verbose=True will automatically disable jit.
```
fabianp, line 182: verbose is an int, but in the doc it's treated as a bool. perhaps we should say `Any value of verbose greater than zero will automatically disable ...`?

jaxopt/_src/lbfgs.py
```diff
@@ -0,0 +1,312 @@
...
+@dataclass(eq=False)
+class LBFGS(base.IterativeSolver):
+  """LBFGS solver.
+
+  Attributes:
+    fun: a smooth function of the form ``fun(x, *args, **kwargs)``.
+    has_aux: whether function fun outputs one (False) or more values (True).
+      When True it will be assumed by default that fun(...)[0] is the objective.
+
+    maxiter: maximum number of proximal gradient descent iterations.
```
fabianp, line 161: spacing is strange here. why are some parameters clustered (sometime there's a blank line between attributes but not always)?

fabianp (approved):
only very cosmetic comments. Thanks!

jaxopt/_src/lbfgs.py
shoyer: The main difference is that the scans inside `inv_hessian_product_leaf` would need to be over a loop index (something like `(time_index - jnp.arange(history_size)[::-1]) % history_size`) with `history` passed in via a closure.

docs/unconstrained.rst
mblondel: Yes it was to show how to instantiate other solvers. I added a comment to explain why.

jaxopt/_src/lbfgs.py
mblondel: I grouped options by "category" because I think very long lists of attributes without any line spacing are very difficult to read.

jaxopt/_src/lbfgs.py
mblondel: Ok switched to boolean for now as I don't need a verbosity level at the moment.

jaxopt/_src/lbfgs.py
mblondel: That's a good idea! I'm kind of sad to remove the scan-based loop as it was pretty elegant but I agree that avoiding memory copies could be very important, especially when working with huge neural networks. I'll tackle that as a separate PR.

jaxopt/_src/lbfgs.py
fabianp: sounds good

fabianp:
🎉 thanks @mblondel ! It's great to see this landed in jaxopt

PR #132, opened 2021-12-16
use epochs instead of max_iter in vision examples.

fabianp:
This change was already implemented for flax_resnet.py, doing it
in the other examples for consistency.


Algue-Rythme:
LGTM !

mblondel:
Thanks both!

PR #143, opened 2021-12-29
customize the VJP of custom-root functions to higher order

### LINKED ISSUE #141: Issue with gradients wrt optimality fn parameters through root finding vjp
First of all, thanks a lot for this library! Really useful tools! 
I'm interested in getting at least 2nd order gradients through root finding, and I'm finding an odd behavior that I wanted to report. 

Maybe I'm doing something wrong, but in the following schematic case I silently get the wrong gradients:
```python
def inv_f(x, aux):
  bisec = Bisection(optimality_fun=F, lower=0.0, upper=1., 
                    check_bracket=False, unroll=True)
  return bisec.run(aux=aux).params

# Here I extract the value part of the vjp, but the grad part also gives wrong results
test_fn = lambda aux: jax.value_and_grad(inv_f)(0.5, aux)[0] 

jax.grad(test_fn)(1.) # Returns 0 instead of the expected gradients
```
Here I'm only trying to get gradients of the value returned by `jax.value_and_grad`, but the gradients of the gradients returned by `jax.value_and_grad` are also wrong (but not as obvious). 

I made a small **demo notebook that reproduces this issue [here](https://colab.research.google.com/drive/1D5FHJJFyXmFuCS45jUjHJvxIQUhBemYy?usp=sharing)**. 

As a reference I've also implemented my own implicit gradients, bypassing the jaxopt ones, and they seem to give me the correct answer.

Reading the source code of jaxopt, it is not immediatly obvious to me why this doesn't work... Sorry I couldn't directly suggest a PR, but I hope this report is still useful (and that I'm not just using jaxopt wrong).

froystig:
When a solver is sent through `custom_root`, we customize its derivative (via `jax.custom_vjp`) to make use of the implicit AD machinery. This change customizes all higher-order derivatives similarly.

Fixes #141 and adds a test for it.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,12 @@ Changelog
 Version 0.3
 -----------
 
+Bug fixes and enhancements
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- `Support implicit AD in higher-order differentiation
+  <https://github.com/google/jaxopt/pull/143>`_.
+
 Version 0.2
 -----------
 
--- a/jaxopt/_src/implicit_diff.py
+++ b/jaxopt/_src/implicit_diff.py
@@ -201,9 +201,13 @@ def _custom_root(solver_fun, optimality_fun, solve, has_aux,
     reference_signature = inspect.signature(fun)
 
   def make_custom_vjp_solver_fun(solver_fun, kwarg_keys):
-    def solver_fun_fwd(*flat_args):
+    @jax.custom_vjp
+    def solver_fun_flat(*flat_args):
       args, kwargs = _extract_kwargs(kwarg_keys, flat_args)
-      res = solver_fun(*args, **kwargs)
+      return solver_fun(*args, **kwargs)
+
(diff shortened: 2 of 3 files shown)


froystig:
This revealed itself in #141 because the example there is based on bisection. Bisection is a solver whose implementation has a different automatic derivative (namely zero) than the one we intend to have, and that we supply via implicit AD. With two applications of AD, the second derivative acted on the solver implementation directly rather than via jaxopt's implicit AD, and so we saw zeros.

This contrasts with other solvers, such as the ridge regression solver in `implicit_diff_test.py`, where AD of the implementation numerically matches the one defined by `custom_root`. Prior to this change, jaxopt's implicit AD wouldn't have been skipped on second application as well, but we may not have noticed it so easily.

mblondel (approved):
Thanks a lot @froystig! LGTM

Could you add a [changelog](https://github.com/google/jaxopt/blob/main/docs/changelog.rst) entry?

PR #146, opened 2022-01-12
Expose AutoOrBoolean.

mblondel:
This is useful for users who want to write their own solver class.


fabianp (approved):
looks good to me!

PR #147, opened 2022-01-12
Pass x0=init to linear solvers

### LINKED ISSUE #145: Linear solvers don't use the `init` argument to warm start solvers.
As an example, the `init` argument in `solve_gmres` is not passed to the `x0` argument in `jax.scipy.sparse.linalg.gmres`.

```python
def solve_gmres(matvec: Callable,
                b: Any,
                ridge: Optional[float] = None,
                init: Optional[Any] = None,
                tol: float = 1e-5,
                **kwargs) -> Any:
  """Solves ``A x = b`` using gmres.
  Args:
    matvec: product between ``A`` and a vector.
    b: pytree.
    ridge: optional ridge regularization.
    init: optional initialization to be used by gmres.
    **kwargs: additional keyword arguments for solver.
  Returns:
    pytree with same structure as ``b``.
  """
  if ridge is not None:
    matvec = _make_ridge_matvec(matvec, ridge=ridge)
  return jax.scipy.sparse.linalg.gmres(matvec, b, tol=tol, **kwargs)[0]
```

I can take care of this soon.

GeoffNN:
Fixes #145

### DIFF
--- a/jaxopt/_src/linear_solve.py
+++ b/jaxopt/_src/linear_solve.py
@@ -150,7 +150,7 @@ def solve_normal_cg(matvec: Callable,
 
   Ab = _rmatvec(matvec, b)
 
-  return jax.scipy.sparse.linalg.cg(_matvec, Ab, **kwargs)[0]
+  return jax.scipy.sparse.linalg.cg(_matvec, Ab, x0=init, **kwargs)[0]
 
 
 def solve_gmres(matvec: Callable,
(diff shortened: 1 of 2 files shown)


mblondel (approved):
LGTM, thanks!

Maybe, in a separate PR, we can stop forwarding all kwargs. The idea is that we don't want to be too tied to the underlying implementation. If some additional options are needed, we need to explicitly add them to the function signature.

GeoffNN:
Got it. There was a shape error in the IterativeRefinementTest.

PR #148, opened 2022-01-13
Add gauss-newton nonlinear least squares optimization 

amir-saadat:
This is an implementation of gauss-newton algorithm, the simple version of nonlinear optimization algorithms.
I have had a couple of iterations with mblondel and applied some comments on applying jvp and vjp to determine the hessian and gradient wrt params. Also using solve_cg to find the solution to system of linear equations instead of direct inv.

### DIFF
--- a/jaxopt/_src/gauss_newton.py
+++ b/jaxopt/_src/gauss_newton.py
@@ -1,0 +1,156 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""GaussNewton algorithm in JAX."""
+
+from typing import Any
+from typing import Callable
+from typing import NamedTuple
+from typing import Optional
+
+from dataclasses import dataclass
+
+import jax
+import jax.numpy as jnp
+
+from jaxopt._src import base
+from jaxopt._src import linear_solve
+from jaxopt.tree_util import tree_l2_norm, tree_sub
+
+
+class GaussNewtonState(NamedTuple):
+  """Named tuple containing state information."""
+  iter_num: int
+  residual: Any
+  delta: Any
+  error: float
+  gradient: Any
+  aux: Optional[Any] = None
+
+@dataclass(eq=False)
+class GaussNewton(base.IterativeSolver):
+  """Gauss-Newton nonlinear least-squares solver.
+
+  This solver finds the optimal parameters via a least-squares optimization
+
+    Given the residual function f(x): R^m -> R^n, `gauss_newton` finds a
+    local minimum of the cost function F(x):
+
+    ```
+    argmin_x F(x) = 0.5 * sum(f_i(x)**2), i = 0, ..., m - 1
+    f(x) = func(x, *args, **kwargs)
+    ```
+
+  Attributes:
+    residual_fun: a smooth function of the form ``residual_fun(x, *args, **kwargs)``.
+    maxiter: maximum number of iterations.
+    tol: tolerance.
+    implicit_diff: whether to enable implicit diff or autodiff of unrolled
+      iterations.
+    implicit_diff_solve: the linear system solver to use.
+    verbose: whether to print error on every iteration or not.
+      Warning: verbose=True will automatically disable jit.
+    jit: whether to JIT-compile the bisection loop (default: "auto").
+    unroll: whether to unroll the bisection loop (default: "auto").
+  """
+  residual_fun: Callable
+  maxiter: int = 30
+  tol: float = 1e-5
+  verbose: bool = False
+  implicit_diff: bool = True
+  implicit_diff_solve: Optional[Callable] = None
+  has_aux: bool = False
+  jit: base.AutoOrBoolean = "auto"
+  unroll: base.AutoOrBoolean = "auto"
+
+  def init_state(self,
+                 init_params: Any,
+                 *args,
+                 **kwargs) -> GaussNewtonState:
+    """Initialize the solver state.
+
+    Args:
+      init_params: pytree containing the initial parameters.
+      *args: additional positional arguments to be passed to ``residual_fun``.
+      **kwargs: additional keyword arguments to be passed to ``residual_fun``.
+    Returns:
+      state
+    """
+    # Compute actual values of state variables at init_param
+    residual, aux = self._fun_with_aux(init_params, *args, **kwargs)
+    matvec = lambda v: self._jtj_op(init_params, v, *args, **kwargs)
+    gradient = self._jt_op(init_params, residual, *args, **kwargs)
+
+    return GaussNewtonState(iter_num=jnp.asarray(0),
+                            error=jnp.asarray(jnp.inf),
+                            residual=residual,
+                            delta=init_params,
+                            gradient=gradient,
+                            aux=aux)
+
+  def update(self,
+             params,
+             state: NamedTuple,
+             *args,
+             **kwargs) -> base.OptStep:
+    """Performs one iteration of the least-squares solver.
+
+    Args:
+      params: pytree containing the parameters.
+      state: named tuple containing the solver state.
+    Returns:
+      (params, state)
+    """
+    residual, aux = self._fun_with_aux(params, *args, **kwargs)
+
+    matvec = lambda v: self._jtj_op(params, v, *args, **kwargs)
+    gradient = self._jt_op(params, residual, *args, **kwargs)
+
+    delta_params = linear_solve.solve_cg(matvec, gradient)
+    params = tree_sub(params, delta_params)
+
+    state = GaussNewtonState(iter_num=state.iter_num + 1,
+                             error=tree_l2_norm(delta_params),
+                             residual=residual,
+                             delta=delta_params,
+                             gradient=gradient,
+                             aux=aux)
+
+    return base.OptStep(params=params, state=state)
+
+  def __post_init__(self):
+
+    if self.has_aux:
+      self._fun = lambda *a, **kw: self.residual_fun(*a, **kw)[0]
+      self._fun_with_aux = self.fun
+    else:
+      self._fun = self.residual_fun
+      self._fun_with_aux = lambda *a, **kw: (self.residual_fun(*a, **kw),
+                                             None)
+    # We need this definition in the base solver run function
+    def optimality_fun(params, *args, **kwargs):
+      residual = self._fun(params, *args, **kwargs)
+      return self._jt_op(params, residual, *args, **kwargs)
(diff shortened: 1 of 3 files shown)


jaxopt/_src/gauss_newton.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+class GaussNewtonState(NamedTuple):
+  """Named tuple containing state information."""
+  iter_num: int
+  residual: Any
+  delta: Any
+  error: float
+  gradient: Any
+  jtj: Optional[Any] = None
```
mblondel, line 40: Why do you store `residual`, `delta`, `gradient` and `jtj`? Are they used in underlying applications?

jaxopt/_src/gauss_newton.py
```diff
@@ -0,0 +1,165 @@
...
+                             aux=aux)
+
+    return base.OptStep(params=params, state=state)
+
+  def run(self,
+          init_params: Optional[Any] = None,
+          *args,
+          **kwargs) -> base.OptStep:
+    # We override run in order to set init_params=None by default.
+    return super().run(init_params, *args, **kwargs)
```
mblondel, line 140: I would be in favor of making `init_params` mandatory. This way, we don't even need to override `run`.

jaxopt/_src/gauss_newton.py
```diff
@@ -0,0 +1,165 @@
...
+      **kwargs: additional keyword arguments to be passed to ``fun``.
+    Returns:
+      state
+    """
+
+    return GaussNewtonState(iter_num=jnp.asarray(0),
+                            error=jnp.asarray(jnp.inf),
+                            residual=jnp.asarray(jnp.inf),
+                            delta=jnp.asarray(jnp.inf),
+                            gradient=jnp.asarray(jnp.inf),
```
mblondel, line 97: Note that attributes `residual`, `delta`, `gradient` and `jtj` are scalars when returned by `init_state` but are arrays or pytrees when returned by `update`. This will trigger a jit recompilation because the state will change signature. So if you really want to keep these attributes, it may be better to compute their actual values at `init_params` and return that instead.

jaxopt/_src/gauss_newton.py
```diff
@@ -0,0 +1,165 @@
...
+    implicit_diff_solve: the linear system solver to use.
+    verbose: whether to print error on every iteration or not.
+      Warning: verbose=True will automatically disable jit.
+    jit: whether to JIT-compile the bisection loop (default: "auto").
+    unroll: whether to unroll the bisection loop (default: "auto").
+  """
+  residual_fun: Callable
+  maxiter: int = 30
+  tol: float = 1e-5
+  check_bracket: bool = True
```
mblondel, line 72: Not used

jaxopt/_src/gauss_newton.py
```diff
@@ -0,0 +1,165 @@
...
+from typing import Optional
+
+from dataclasses import dataclass
+
+import jax
+import jax.numpy as jnp
+
+from jaxopt._src import base
+from jaxopt._src import implicit_diff as idf
+from jaxopt._src import loop
```
mblondel, line 29: Not used

jaxopt/_src/gauss_newton.py
```diff
@@ -0,0 +1,165 @@
...
+    else:
+      params = jnp.asarray(params, float)
+
+    residual, aux = self._fun_with_aux(params, *args, **kwargs)
+
+    matvec = lambda v: self._jtj(params, v, *args, **kwargs)
+    gradient = self._gradient(params, residual, *args, **kwargs)
+
+    delta_params = linear_solve.solve_cg(matvec, gradient)
+    params -= delta_params
```
mblondel, line 124: You can use `tree_sub` here.

jaxopt/_src/gauss_newton.py
```diff
@@ -0,0 +1,165 @@
...
+
+    state = GaussNewtonState(iter_num=state.iter_num + 1,
+                             error=jnp.linalg.norm(delta_params),
```
mblondel, line 127: And `tree_l2_norm` here

jaxopt/_src/gauss_newton.py
```diff
@@ -0,0 +1,165 @@
...
+  def run(self,
+          init_params: Optional[Any] = None,
+          *args,
+          **kwargs) -> base.OptStep:
+    # We override run in order to set init_params=None by default.
+    return super().run(init_params, *args, **kwargs)
+
+  def __post_init__(self):
+    # We need this definition in the base solver run function
+    self.optimality_fun = self.residual_fun
```
mblondel, line 144: `optimality_fun` should contain the gradient of `0.5 ||residual_fun(params)||^2`, i.e. `optimality_fun(params) = J.T residual_fun(params)`. So if I'm not wrong it should be:
```python
def optimality_fun(params, *args, **kwargs):
  residual = self.residual_fun(params, *args, **kwargs)
  return self._gradient(self, params, residual, *args, **kwargs)
```

jaxopt/_src/gauss_newton.py
```diff
@@ -0,0 +1,165 @@
...
+    else:
+      self._fun = self.residual_fun
+      self._fun_with_aux = lambda *a, **kw: (self.residual_fun(*a, **kw),
+                                             None)
+
+  def _fun_with_args(self, *args, **kwargs):
+    fun_with_args = lambda p: self._fun(p, *args, **kwargs)
+    return self._fun_with_args
+
+  def _jtj(self, params, vec, *args, **kwargs):
```
mblondel, line 158: Maybe add the docstring `"""Product with J.T J"""`?

mblondel (commented):
Thanks for your contribution @amir-saadat! I left some comments and questions.

jaxopt/_src/gauss_newton.py
amir-saadat: Good catch, thanks for reminding.

jaxopt/_src/gauss_newton.py
amir-saadat: Fixed. Thanks.

jaxopt/_src/gauss_newton.py
amir-saadat: Yes, that's a great point. This is essentially the RHS of the main equation we are solving to advance the params in each step.

jaxopt/_src/gauss_newton.py
amir-saadat: I implemented your suggestion, very good point, thanks.

jaxopt/_src/gauss_newton.py
amir-saadat: Applied.

jaxopt/_src/gauss_newton.py
amir-saadat: Yes, exactly, they are useful for interpreting the results for the underlying application. for example, if needed, eigenvalues of JT.J can be evaluated. For LM specifically, grad and delta can be useful to determine what cause the algorithm to stop/converge.

amir-saadat:
Thanks @mblondel for the first round of review. They were very helpful and I think I have implemented them now.

jaxopt/_src/gauss_newton.py
```diff
@@ -0,0 +1,170 @@
...
+    Given the residual function f(x): R^m -> R^n, `gauss_newton` finds a
+    local minimum of the cost function F(x):
+
+    ```
+    argmin_x F(x) = 0.5 * sum(f_i(x)**2), i = 0, ..., m - 1
+    f(x) = func(x, *args, **kwargs)
+    ```
+
+  Attributes:
+    residual_fun: a smooth function of the form ``fun(x, *args, **kwargs)``.
```
mblondel, line 57:
```suggestion
    residual_fun: a smooth function of the form ``residual_fun(x, *args, **kwargs)``.
```

jaxopt/_src/gauss_newton.py
```diff
@@ -0,0 +1,170 @@
...
+    maxiter: maximum number of iterations.
+    tol: tolerance.
+    check_bracket: whether to check correctness of the bracketing interval.
```
mblondel, line 60: Remove

jaxopt/_src/gauss_newton.py
```diff
@@ -0,0 +1,170 @@
...
+
+  def init_state(self,
+                 init_params: Any,
+                 *args,
+                 **kwargs) -> GaussNewtonState:
+    """Initialize the solver state.
+
+    Args:
+      init_params: pytree containing the initial parameters.
+      *args: additional positional arguments to be passed to ``fun``.
```
mblondel, line 85:
```suggestion
      *args: additional positional arguments to be passed to ``residual_fun``.
```

mblondel: And a few other occurrences.

jaxopt/_src/gauss_newton.py
```diff
@@ -0,0 +1,170 @@
...
+      **kwargs: additional keyword arguments to be passed to ``fun``.
+    Returns:
+      state
+    """
+
+    # Compute actual values of state variables at init_param
+    residual, aux = self._fun_with_aux(init_params, *args, **kwargs)
+    matvec = lambda v: self._jtj_op(init_params, v, *args, **kwargs)
+    gradient = self._jt_op(init_params, residual, *args, **kwargs)
+    # Only materialize J^T.J if verbose is True for debugging
```
mblondel, line 95: I would rather remove `jtj` from the state if possible. I don't think it's a good idea that the state depends on the `verbose` option. If you really need it, I think I would rather add an option `store_jtj` in the constructor.

mblondel (commented):
Thanks a lot @amir-saadat. I left a few other comments but overall this looks good! 

It would be great to add an implicit diff test, see e.g. [here](https://github.com/google/jaxopt/blob/main/tests/gradient_descent_test.py#L54) for inspiration. For instance in your `_city_temperature_residual_model` function, you could try to differentiate w.r.t. `x` or `y`.

We will eventually also need documentation and potentially an example but those can be tackled in a separate PR. I think it's better to merge this one first.

jaxopt/_src/gauss_newton.py
amir-saadat: I removed `jtj`. As you mentioned, this becomes more like a debugging feature that I have to think how to extract for the optimizer. Eventually, the eignevalues of the matrix might be of interest. I'll get it into a separate PR once i have an idea.

amir-saadat:
Thanks @mblondel for the nice suggestions. I added the unit test `test_enzyme_reaction_implicit_diff`. It sounded non-trivial conceptually, so please double check that I have done the expected process. Also please let me know how `implicit_diff` is supposed to change the behavior. In the current framework, it doesn't seem to change the results.

jaxopt/_src/gauss_newton.py
```diff
@@ -69,6 +68,7 @@ class GaussNewton(base.IterativeSolver):
   maxiter: int = 30
   tol: float = 1e-5
   verbose: bool = False
+  implicit_diff: bool = False
```
mblondel, line 71: We tend to use `True` as the default in other solvers.

mblondel (approved):
The new implicit diff test looks good to me. Thanks a lot!

Could you squash your commits? Then, we'll merge!

jaxopt/_src/gauss_newton.py
amir-saadat: Thanks, applied.

PR #153, opened 2022-01-20
Use circular buffer in LBFGS implementation.

mblondel:
Following @shoyer's idea, I implemented a circular buffer for maintaining the histories, thus avoiding memory copies. All tests are passing. I hope I didn't make any mistake!

### DIFF
--- a/jaxopt/_src/lbfgs.py
+++ b/jaxopt/_src/lbfgs.py
@@ -38,48 +38,56 @@ from jaxopt.tree_util import tree_sum
 from jaxopt.tree_util import tree_l2_norm
 
 
-def ihp_body_right(q, tup):
-  s, y, rho = tup
-  alpha = rho * tree_vdot(s, q)
-  q = tree_add_scalar_mul(q, -alpha, y)  # q = q - alpha * y
-  return q, alpha
-
-
-def ihp_body_left(r, tup):
-  s, y, rho, alpha = tup
-  beta = rho * tree_vdot(y, r)
-  r = tree_add_scalar_mul(r, alpha - beta, s)  # r = r + (alpha - beta) * s
-  return r, beta
-
-
 def inv_hessian_product_leaf(v: jnp.ndarray,
                              s_history: jnp.ndarray,
                              y_history: jnp.ndarray,
                              rho_history: jnp.ndarray,
-                             gamma: float = 1.0):
+                             gamma: float = 1.0,
+                             start: int = 0):
 
-  # Compute right part.
-  q, alpha = jax.lax.scan(ihp_body_right,
-                          v,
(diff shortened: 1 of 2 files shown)


shoyer:
Nice! Have you tried running any micro-benchmarks to compare performance?

soraros:
Maybe something like this is slightly cleaner, as the circular indexing is dealt with in a single place.
```python
def inv_hessian_product_leaf(v: jnp.ndarray,
                             s_history: jnp.ndarray,
                             y_history: jnp.ndarray,
                             rho_history: jnp.ndarray,
                             gamma: float = 1.0,
                             start: int = 0):
  history_size = len(s_history)
  I = (start - 1 - jnp.arange(history_size)) % history_size

  def body_right(r, i):
    alpha = rho_history[i] * jnp.vdot(s_history[i], r)
    r = r - alpha * y_history[i]
    return r, alpha

  r, alpha = lax.scan(body_right, v, I)

  r = r * gamma

  def body_left(r, tup):
    i, alpha = tup
    beta = rho_history[i] * jnp.vdot(y_history[i], r)
    r = r + s_history[i] * (alpha - beta)
    return r, beta

  r, beta = lax.scan(body_left, r, (I, alpha), reverse=True)
  return r
```

mblondel:
@soraros I agree your version is cleaner. I switched to it, thanks a lot!

@shoyer On toy problems, the new version seems slightly faster. BTW, I plan to do a proper benchmark against SciPy's LBFGS in a separate PR.

jaxopt/_src/lbfgs.py
```diff
@@ -38,39 +38,33 @@
...
 
-  # Compute right part.
-  q, alpha = jax.lax.scan(ihp_body_right,
-                          v,
-                          (s_history, y_history, rho_history),
-                          reverse=True)
+  def body_right(r, i):
+    alpha = rho_history[i] * jnp.vdot(s_history[i], r)
+    r = r - alpha * y_history[i]
+    return r, alpha
```
shoyer, line 55: I guess you didn't need to handle pytrees in these helper functions after all?

shoyer (approved):

jaxopt/_src/lbfgs.py
mblondel: I'm not sure why but in the previous version of the code, I was getting errors if I didn't use tree utils, even though `inv_hessian_product_leaf` is supposed to only receive tree leaves... With the new version of the code, all tests passed directly with `jnp` functions so I went with that. Shall I use the tree utils instead?

shoyer: You have test coverage, so if things work with the new code, that's fine by me!

PR #162, opened 2022-02-01
Remove dependence on JaxTestCase.

mblondel:
This PR introduces `JaxoptTestCase` as a replacement for `JaxTestCase`. I tried to copy the minimal amount of code so all tests pass.


fllinares (approved):
Thanks a lot Mathieu, LGTM!

Issue #70, opened 2021-10-13
vmap support in QPs


jecampagne:
Hi, 
I experience some pb with ` projection_polyhedron`

```python
import numpy as np
import matplotlib.pyplot as plt

import jax
import jax.numpy as jnp

import jaxopt
from jaxopt.projection import projection_l2_ball, projection_box, projection_l1_ball, projection_polyhedron

def myproj3(x):
    A = jnp.array([[1.0, 1.0]])
    b = jnp.array([1.0])
    G = jnp.array([[-1.0, 0.0], [0.0, -1.0]])
    h = jnp.array([0.0, 0.0])    
    x = projection_polyhedron(x,hyperparams = (A, b, G, h))
    return x

rng_key = jax.random.PRNGKey(42)
x = jax.random.uniform(rng_key, (5000,2), minval=-3,maxval=3)
p1_x=jax.vmap(myproj3, in_axes=(0,None))(x)
fig, ax = plt.subplots(figsize=(5,5))
ax.scatter(x[:,0],x[:,1],s=0.5)
ax.scatter(p1_x[:,0],p1_x[:,1],s=0.5,c='g')
ax.set_xlabel("X")
ax.set_ylabel("Y")
plot.show()
```

First, I had to install `cvxpy`
#!pip install cvxpy
Then, I got this error
```
TracerArrayConversionError: The numpy.ndarray conversion method __array__() was called on the JAX Tracer object Traced<ShapedArray(float64[2])>with<BatchTrace(level=1/1)>
  with val = DeviceArray([[-2.37103211,  2.33759997],
                          [ 2.76953806, -2.37750394],
                          [-0.87246632,  0.73224625],
                          ...,
                          [ 2.29799773,  2.81894884],
                          [ 2.4022714 ,  0.80693103],
                          [-0.41563116,  2.83898531]], dtype=float64)
       batch_dim = 0
```

Is anyone has an hint? Thanks

jecampagne:
Ok for `vmap`now, I experience another problem but it may be related to the way I have coded. 
I have tried to make the following:
3 lines forming a triangle with A & b matrix/vector
y_0 + y_1 = 1
y_1 = 1
-y_0+y_1 = _1
and with G & h
y_0 >0 & y_1 >0

```python
def myproj3(x):
    A = jnp.array([[1.0, 1.0],[0.0,1.0],[-1.0,1.0]])
    b = jnp.array([1.0,1.0,-1.0])
[...]

PR #163, opened 2022-02-01
Remove QuadraticProg in projections

Algue-Rythme:
## What changed ?

Last steps in the removal of `QuadraticProgramming` class: replacing it by `OSQP` and `EqualityConstrainedQP` in `_src/projections.py`.

By default the function `projection_polyhedron` checks if the polyhedron is empty ([issue encountered by a user](https://github.com/google/jaxopt/issues/70#issuecomment-943701279)). This check is performed by raising an exception. This prevents the code of the projection to be jittable. `OSQP` offers ways to check feasability within `jit` block without raising an exception but by returning an error code in `state.status` instead. This can be left for future work.

## Future work

In future PRs we could wrap projections in classes (instead of functions) to separate differentiable parameters and non differentiable hyper-parameters.

### DIFF
--- a/jaxopt/_src/osqp.py
+++ b/jaxopt/_src/osqp.py
@@ -36,7 +36,15 @@ from jaxopt._src.tree_util import tree_ones_like, tree_zeros_like, tree_where
 from jaxopt._src.tree_util import tree_negative, tree_l2_norm, tree_inf_norm
 from jaxopt._src.linear_operator import DenseLinearOperator, _make_linear_operator
 import jaxopt.linear_solve as linear_solve
-from jaxopt.projection import projection_box
+
+
+# Since jaxopt.projection itself depends on OSQP, we duplicate projection_box to avoid a circular dependency.
+def _clip_safe(x, lower, upper):
+  return jnp.clip(jnp.asarray(x), lower, upper)
+
+def projection_box(x: Any, hyperparams: Tuple) -> Any:
+  lower, upper = hyperparams
+  return tree_map(_clip_safe, x, lower, upper)
 
 
 def _make_osqp_optimality_fun(matvec_Q, matvec_A):
@@ -636,21 +644,24 @@ class BoxOSQP(base.IterativeSolver):
     return base.OptStep(params=sol, state=state)
 
   def run(self,
-          init_params: Any,
-          params_obj: Tuple[Optional[Any], Any],  # Q can be params_Q or None
-          params_eq: Optional[Any],  # A can be params_A or None
-          params_ineq: Tuple[Optional[Any], Any]) -> base.OptStep:
+          init_params: Optional[Any] = None,
+          params_obj: Tuple[Optional[Any], Any] = None,  # Q can be params_Q or None
+          params_eq: Optional[Any] = None,  # A can be params_A or None
+          params_ineq: Tuple[Any, Any] = None) -> base.OptStep:
     """Return primal/dual variables.
 
     Args:
-      init_params: initial KKTSolution (can be None).
+      init_params: (optional) initial KKTSolution.
       params_obj: pair (params_Q, c).
-      params_eq: params_A.
+      params_eq: (optional) params_A.
       params_ineq: pair (l, u).
     """
+    assert params_obj is not None
+    assert params_ineq is not None
+
     if init_params is None:
       init_params = self.init_params(None, params_obj, params_eq, params_ineq)
-
+    
     return super().run(init_params, params_obj, params_eq, params_ineq)
 
   def l2_optimality_error(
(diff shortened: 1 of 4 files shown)


jaxopt/_src/osqp.py
```diff
@@ -36,7 +36,15 @@
 from jaxopt._src.tree_util import tree_negative, tree_l2_norm, tree_inf_norm
 from jaxopt._src.linear_operator import DenseLinearOperator, _make_linear_operator
 import jaxopt.linear_solve as linear_solve
-from jaxopt.projection import projection_box
+
+
+# Since projections depends of OSQP, we need to duplicate the code into osqp.py
```
mblondel, line 41:
```suggestion
# Since jaxopt.projection itself depends on OSQP, we duplicate projection_box to avoid a circular dependency.
```

tests/projection_test.py
```diff
@@ -328,9 +328,9 @@ def test_projection_box_section(self):
     G = jnp.concatenate((-jnp.eye(len(x)), jnp.eye(len(x))))
     h = jnp.concatenate((alpha, beta))
     hyperparams = dict(params_obj=(Q, -x), params_eq=(A, b), params_ineq=(G, h))
-    qp = QuadraticProgramming()
-    p2 = qp.run(**hyperparams).params[0]
-    self.assertArraysAllClose(p, p2, atol=1e-5)
+    osqp = OSQP()
+    p2 = osqp.run(None, **hyperparams).params.primal
```
mblondel, line 332: Same here.

jaxopt/_src/projection.py
```diff
@@ -289,16 +292,22 @@ def projection_polyhedron(x: jnp.ndarray, hyperparams: Tuple) -> jnp.ndarray:
...
   A, b, G, h = hyperparams
-  qp = QuadraticProgramming()
-  I = jnp.eye(len(x))
-  hyperparams = dict(params_obj=(I, -x), params_eq=(A, b), params_ineq=(G, h))
-  return qp.run(**hyperparams).params[0]
+  matvec_Q = lambda _, vec: vec
+  osqp = OSQP(matvec_Q=matvec_Q, check_primal_dual_infeasability=check_feasible)
+  # check feasability by default; currently there is no way to return the info to the user inside @jit.
+  hyperparams = dict(params_obj=(None, -x), params_eq=(A, b), params_ineq=(G, h))
+  kkt_sol, state = osqp.run(init_params=None, **hyperparams)
```
mblondel, line 307: Same here.

jaxopt/_src/projection.py
```diff
@@ -269,15 +270,17 @@ def projection_affine_set(x: jnp.ndarray, hyperparams: Tuple) -> jnp.ndarray:
...
+  # TODO: support matvec for A
   A, b = hyperparams
-  qp = QuadraticProgramming()
-  I = jnp.eye(len(x))
-  hyperparams = dict(params_obj=(I, -x), params_eq=(A, b), params_ineq=None)
-  return qp.run(**hyperparams).params[0]
+  matvec_Q = lambda _, vec: vec
+  osqp = EqualityConstrainedQP(matvec_Q=matvec_Q)
+  hyperparams = dict(params_obj=(None, -x), params_eq=(A, b))
+  kkt_sol = osqp.run(init_params=None, **hyperparams).params
```
mblondel, line 278: It would be nice if `None` is the default and doesn't need to be provided.

mblondel (commented):
Thanks a lot @Algue-Rythme! My main comment is that it would be nice if`init_params=None` is the default in all QP classes.

jaxopt/_src/projection.py
```diff
@@ -289,16 +292,22 @@ def projection_polyhedron(x: jnp.ndarray, hyperparams: Tuple) -> jnp.ndarray:
     x: pytree to project.
     hyperparams: tuple ``hyperparams = (A, b, G, h)``, where ``A`` is a matrix,
       ``b`` is a vector, ``G`` is a matrix and ``h`` is a vector.
+    check_feasability: if True (default: True) check the non emptyness of the polyhedron,
```
mblondel, line 295:
```suggestion
    check_feasible: if True (default: True) check the non emptyness of the polyhedron,
```

Algue-Rythme:
Since `run()` must share the signature of all solvers (and of optimality_fun) the init_params has to be the first parameter. By forcing its default value to `None` we must chose a default value for other parameters `params_obj,params_eq,params_ineq` (because arguments with non defaul value cannot follow arguments with default value). So I decided to assign `None` by default for those, even though it does not always make sense. For this reason, the value `None` will remain undocumented and not present in signature.

Issue #174, opened 2022-02-19
`BacktrackingLineSearch` created in LBFGS does not use same values for `jit` and `unroll`


ianwilliamson:
I would have expected the line search in LBFGS to use the same values for its `jit` and `unroll` as the enclosing LBFGS optimizer, however this doesn't seem to be the case:

https://github.com/google/jaxopt/blob/0ff6be8094aacb3bf5472a41d780e3f56fc8e0f8/jaxopt/_src/lbfgs.py#L258-L262

Is this intentional?

mblondel:
Fixed by #175.

PR #175, opened 2022-02-19
Update line search in LBFGS to use jit and unroll settings from LBFGS

ianwilliamson:
Addresses #174.

### DIFF
--- a/jaxopt/_src/lbfgs.py
+++ b/jaxopt/_src/lbfgs.py
@@ -259,7 +259,9 @@ class LBFGS(base.IterativeSolver):
                                 value_and_grad=True,
                                 maxiter=self.maxls,
                                 decrease_factor=self.decrease_factor,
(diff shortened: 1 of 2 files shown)


mblondel:
Thanks a lot for the contribution @ianwilliamson! I presume you changed this because jit compilation was failing? In this case, it would be useful to add LBFGS in this [common test](https://github.com/google/jaxopt/blob/main/tests/common_test.py#L35) (which I forgot to do).

ianwilliamson:
> I presume you changed this because jit compilation was failing?

Yes, that's right. I was using a function that didn't support jit.

> In this case, it would be useful to add LBFGS in this [common test](https://github.com/google/jaxopt/blob/main/tests/common_test.py#L35) (which I forgot to do).

Done.

mblondel (approved):
LGTM, thanks a lot!

PR #178, opened 2022-02-25
Distributed Flax ImageNet example

fllinares:
The purpose of this example is to illustrate how JAXopt solvers can be easily used for distributed training thanks to `jax.pjit`. In this case, we begin by implementing data parallel training of a `ResNet50` model on the `ImageNet` datasets as a fork of [Flax's official ImageNet example](https://github.com/google/flax/tree/main/examples/imagenet). General aspects to pay attention to include:
+ How auxiliary information (e.g. Flax mutables, model outputs from train metrics, etc) can be extracted from `loss_fun` using the `state.aux` field of JAXopt's optimizer state.
+ How `jax.pjit` can be used to easily port single-device training loops to distributed training loops.

Future examples will cover data and model parallel training using `jax.pjit` as well as support for `jax.pmap`-based distributed training.

### DIFF
--- a/examples/deep_learning/distributed_flax_imagenet.py
+++ b/examples/deep_learning/distributed_flax_imagenet.py
@@ -1,0 +1,867 @@
+# Copyright 2022 Google LLC.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+====================================
+SPMD ResNet example with Flax and JAXopt.
+====================================
+The purpose of this example is to illustrate how JAXopt solvers can be easily
+used for distributed training thanks to `jax.pjit`. In this case, we begin by
+implementing data parallel training of a ResNet50 model on the ImageNet dataset
+as a fork of Flax's official ImageNet example. General aspects to pay attention
+to include:
+ + How auxiliary information (e.g. Flax mutables, model outputs from train
+     metrics, etc) can be extracted from `loss_fun` using the `state.aux` field
+     of JAXopt's optimizer state.
+ + How `jax.pjit` can be used to easily port single-device training loops to
+     distributed training loops.
+
+Running on Google Cloud TPU:
+
+1. Follow the instructions in Flax's official ImageNet example to set a single
+   VM with 8 TPUs (`--accelerator_type v3-8`).
+2. Likewise, follow the instructions in Flax's official ImageNet example to
+   prepare the ImageNet dataset and ensure the `TFDS_DATA_DIR` environment
+   variable has been set appropriately.
+
+You may finally run the example as
+```
+python3 distributed_flax_imagenet.py --workdir=$HOME/spmd_flax_imagenet
+```
+
+NOTES: this example supports TPU pod slices (e.g. `--accelerator_type v3-32`) as
+well as hosts with one or more GPUs attached. However, CPU-only execution is not
+yet supported.
+"""
+
+
+import functools
+import itertools
+import os
+import time
+from typing import Any, Callable, Iterator, Mapping, NamedTuple, Optional, Sequence, Type, Tuple, Union
+
+from absl import app
+from absl import flags
+from absl import logging
+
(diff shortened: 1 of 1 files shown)


examples/deep_learning/distributed_flax_imagenet.py
```diff
@@ -0,0 +1,839 @@
...
+
+import jax
+from jax import numpy as jnp
+from jax import random
+from jax.experimental.maps import Mesh
+from jax.experimental.pjit import PartitionSpec
+from jax.experimental.pjit import pjit
+
+import jaxopt
+from jaxopt import loss as jaxopt_losses
```
mblondel, line 48: I would just use `jaxopt.loss` directly where you need it instead.

examples/deep_learning/distributed_flax_imagenet.py
```diff
@@ -0,0 +1,839 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+SPMD ResNet example with Flax and JAXopt.
+====================================
```
mblondel, line 17: I would add a short description, such as the one in the pull-request description. Briefly explaining concepts such as SPMD and pjit would be useful too. Also maybe add some instructions on how to run it?

examples/deep_learning/distributed_flax_imagenet.py
```diff
@@ -0,0 +1,839 @@
...
+) -> ArrayTree:
+  """Initializes the state of jaxopt.OptaxSolver."""
+  # "Default" JAXopt initial optimizer state.
+  state = solver.init_state(init_params)
+
+  # To prevent `train_step` from being compiled twice, we must ensure all its
+  # input arguments have the same shape and dtype in all calls. To this end,
+  # we will
+  #   1) Initialize `state.aux` with a Pytree of the right shape and dtype.
+  #   2) Ensure that `state.value` and `state.error` are strongly typed.
```
mblondel, line 642: This requires a lot of low-level knowledge from the user. In the future, it would be nice if we can find a way to avoid this. CC @froystig

examples/deep_learning/distributed_flax_imagenet.py
```diff
@@ -0,0 +1,839 @@
...
+) -> Tuple[ArrayTree, ArrayTree, TrainMetrics]:
+  """Performs a single training step."""
+  # Retrieves Flax mutables from previous step, stored in `state.aux`.
+  batch_stats = state.aux.batch_stats
+  # Computes updated model parameters and optimizer state.
+  params, state = solver.update(
+      params=params,
+      state=state,
+      batch_stats=batch_stats,
+      batch=batch,
```
mblondel, line 504: In `run_iterator`, we call it `data`. Maybe `batch` would have been a better name in retrospect, as it conveys the idea of "subsample"? Since not so many people use `run_iterator` yet, I believe, maybe it's not too late to change? I believe these small naming consistencies are very important to provide a nice user experience.

mblondel (approved):
Thank you @fllinares for this amazing work, putting all these pieces together.

examples/deep_learning/distributed_flax_imagenet.py
fllinares: The `initialize_solver` and `initialize_metrics` functions could be somewhat less "intimidating" if I had hard-coded the initialization, e.g.,
```
init_aux = LossFnAux(batch_stats=init_batch_stats, logits=jnp.zeros([batch_size, NUM_CLASSES], dtype=jnp.bfloat16))
```
However, I wanted to use this example to see if we could infer the initialization of `state.aux` and `metrics` automatically instead of asking the user to specify it. These functions show we can, but there might certainly be better ways to accomplish the same.

Going forward, I anticipate these functions (or an improved version of them) could be part of the future `solver.run` / `solver.run_iterator` API. But I do wonder if we should make them somehow accessible to users writing their own manual training loops.

examples/deep_learning/distributed_flax_imagenet.py
fllinares: SGTM, done!

examples/deep_learning/distributed_flax_imagenet.py
fllinares: Done! Perhaps in the future we should add individual markdown instructions for the "heavier" examples to cover things like setting up the GCP VM instances?

examples/deep_learning/distributed_flax_imagenet.py
fllinares: Yes, that's a good point. I "inherited"  `batch` from the Flax ImageNet example, but certainly +1 for consistency with `run_iterator`. I have no strong preference myself on whether to use `data` or `batch`.

PR #180, opened 2022-02-26
Add support for leaf cond pytrees in `tree_util.tree_where`.

fllinares:
Generalizes `tree_util.tree_where` by allowing any of its arguments to be a leaf pytree (i.e. single array), including `cond`. The function `broadcast_pytree` was factored out of `tree_where`, as it could presumably be used in combination with other `tree_x`methods in `tree_util`.

### DIFF
--- a/jaxopt/_src/tree_util.py
+++ b/jaxopt/_src/tree_util.py
@@ -15,6 +15,7 @@
 """Tree utilities."""
 
 import functools
+import itertools
 import operator
 
 import jax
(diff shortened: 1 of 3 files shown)


mblondel:
LGTM on my side but @Algue-Rythme may want to take a look, as the author of the original code.

Algue-Rythme:
I did not run the code so I am not sure; but it feels like `is_scalar_i` returns True for every array/tensor (not only scalars, i.e 0-D arrays), so the name of the variable is a bit misleading. To be fair, I did the same mistake in `tree_where`.

fllinares:
Thanks a lot @Algue-Rythme for the feedback!

I fully agree that, if what I implemented is indeed what was intended, I should have used a different name. What about `is_leaf_i` or `is_leaf_pytree_i`?

Just to double-check, naming of the variable aside, do you think that `broadcast_pytree` captures well the logic for the edge cases of `a` and `b` being individual arrays that you had in mind in `tree_where`?

mblondel (approved):

PR #185, opened 2022-03-09
Fixed solve_normal_cg when matvec is "nonsquare".

### LINKED ISSUE #183: Incompatible shape in solve_normal_cg
When `A.shape = (N, P)` for `N != P`, I run into shape errors when trying to use `solve_normal_cg` for fitting the normal equations.

I have a small reproducible example below for `N > P`, but the error holds for when `P > N`.

```
import jax.numpy as jnp
import numpy as np
N = 1000
P = 3
prob = np.random.uniform(0.01, 0.5, size=P)
h2g = 0.1
X = np.random.binomial(2, p=prob, size=(N, P))
b = np.random.normal(size=(P)) * np.sqrt(h2g / P)
y = X @ b + np.sqrt(1 - h2g) * np.random.normal(size=(N,))

import jaxopt as jopt
jopt.linear_solve.solve_normal_cg(lambda x: jnp.dot(X, x), y)
WARNING:absl:No GPU/TPU found, falling back to CPU. (Set TF_CPP_MIN_LOG_LEVEL=0 and rerun for more info.)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [11], in <module>
----> 1 jopt.linear_solve.solve_normal_cg(lambda x: jnp.dot(X, x), y)

File ~/miniconda3/lib/python3.9/site-packages/jaxopt/_src/linear_solve.py:151, in solve_normal_cg(matvec, b, ridge, init, **kwargs)
    148 if ridge is not None:
    149   _matvec = _make_ridge_matvec(_matvec, ridge=ridge)
--> 151 Ab = _rmatvec(matvec, b)
    153 return jax.scipy.sparse.linalg.cg(_matvec, Ab, x0=init, **kwargs)[0]

File ~/miniconda3/lib/python3.9/site-packages/jaxopt/_src/linear_solve.py:114, in _rmatvec(matvec, x)
    112 def _rmatvec(matvec, x):
    113   """Computes A^T x, from matvec(x) = A x, where A is square."""
--> 114   transpose = jax.linear_transpose(matvec, x)
    115   return transpose(x)[0]

File ~/miniconda3/lib/python3.9/site-packages/jax/_src/api.py:2211, in linear_transpose(fun, reduce_axes, *primals)
   2208 in_dtypes = map(dtypes.dtype, in_avals)
   2210 in_pvals = map(pe.PartialVal.unknown, in_avals)
-> 2211 jaxpr, out_pvals, consts = pe.trace_to_jaxpr(flat_fun, in_pvals,
   2212                                              instantiate=True)
   2213 out_avals, _ = unzip2(out_pvals)
   2214 out_dtypes = map(dtypes.dtype, out_avals)

File ~/miniconda3/lib/python3.9/site-packages/jax/interpreters/partial_eval.py:505, in trace_to_jaxpr(fun, pvals, instantiate)
    503 with core.new_main(JaxprTrace) as main:
    504   fun = trace_to_subjaxpr(fun, main, instantiate)
--> 505   jaxpr, (out_pvals, consts, env) = fun.call_wrapped(pvals)
    506   assert not env
    507   del main, fun, env

File ~/miniconda3/lib/python3.9/site-packages/jax/linear_util.py:166, in WrappedFun.call_wrapped(self, *args, **kwargs)
    163 gen = gen_static_args = out_store = None
    165 try:
--> 166   ans = self.f(*args, **dict(self.params, **kwargs))
    167 except:
    168   # Some transformations yield from inside context managers, so we have to
    169   # interrupt them before reraising the exception. Otherwise they will only
    170   # get garbage-collected at some later time, running their cleanup tasks only
    171   # after this exception is handled, which can corrupt the global state.
    172   while stack:

Input In [11], in <lambda>(x)
----> 1 jopt.linear_solve.solve_normal_cg(lambda x: jnp.dot(X, x), y)

File ~/miniconda3/lib/python3.9/site-packages/jax/_src/numpy/lax_numpy.py:4196, in dot(a, b, precision)
   4194   return lax.mul(a, b)
   4195 if _max(a_ndim, b_ndim) <= 2:
-> 4196   return lax.dot(a, b, precision=precision)
   4198 if b_ndim == 1:
   4199   contract_dims = ((a_ndim - 1,), (0,))

File ~/miniconda3/lib/python3.9/site-packages/jax/_src/lax/lax.py:667, in dot(lhs, rhs, precision, preferred_element_type)
    664   return dot_general(lhs, rhs, (((lhs.ndim - 1,), (0,)), ((), ())),
    665                      precision=precision, preferred_element_type=preferred_element_type)
    666 else:
--> 667   raise TypeError("Incompatible shapes for dot: got {} and {}.".format(
    668       lhs.shape, rhs.shape))

TypeError: Incompatible shapes for dot: got (1000, 3) and (1000,).
```

mblondel:
Fixes #183

### DIFF
--- a/jaxopt/_src/linear_solve.py
+++ b/jaxopt/_src/linear_solve.py
@@ -109,10 +109,9 @@ def solve_cg(matvec: Callable,
   return jax.scipy.sparse.linalg.cg(matvec, b, x0=init, **kwargs)[0]
 
 
-def _rmatvec(matvec, x):
-  """Computes A^T x, from matvec(x) = A x, where A is square."""
+def _make_rmatvec(matvec, x):
   transpose = jax.linear_transpose(matvec, x)
-  return transpose(x)[0]
+  return lambda y: transpose(y)[0]
 
 
 def _normal_matvec(matvec, x):
@@ -141,16 +140,28 @@ def solve_normal_cg(matvec: Callable,
   Returns:
     pytree with same structure as ``b``.
   """
-  def _matvec(x):
-    """Computes A^T A x."""
+  if init is None:
+    example_x = b  # This assumes that matvec is a square linear operator.
+  else:
+    example_x = init
+
+  try:
+    rmatvec = _make_rmatvec(matvec, example_x)
+  except TypeError:
+    raise TypeError("The initialization `init` of solve_normal_cg is "
+                    "compulsory when `matvec` is nonsquare. It should "
+                    "have the same pytree structure as a solution. "
+                    "Typically, a pytree filled with zeros should work.")
+
+  def normal_matvec(x):
     return _normal_matvec(matvec, x)
 
   if ridge is not None:
-    _matvec = _make_ridge_matvec(_matvec, ridge=ridge)
+    normal_matvec = _make_ridge_matvec(normal_matvec, ridge=ridge)
 
-  Ab = _rmatvec(matvec, b)
+  Ab = rmatvec(b)  # A.T b
 
-  return jax.scipy.sparse.linalg.cg(_matvec, Ab, x0=init, **kwargs)[0]
+  return jax.scipy.sparse.linalg.cg(normal_matvec, Ab, x0=init, **kwargs)[0]
 
 
 def solve_gmres(matvec: Callable,
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,15 @@
 Changelog
 =========
 
+Main branch
+-----------
+
+Bug fixes and enhancements
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- Fixed ``solve_normal_cg`` when the linear operator is "nonsquare" (does not map to a space of same dimension),
+  by Mathieu Blondel.
+
 Version 0.3.1.
 --------------
 
--- a/tests/linear_solve_test.py
+++ b/tests/linear_solve_test.py
@@ -23,6 +23,7 @@ from jaxopt._src import test_util
 
 import numpy as onp
 
+
 class LinearSolveTest(test_util.JaxoptTestCase):
 
   def test_materialize_array(self):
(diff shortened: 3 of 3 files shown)


jaxopt/_src/linear_solve.py
```diff
@@ -141,16 +153,34 @@ def solve_normal_cg(matvec: Callable,
...
+  try:
+    rmatvec = _make_rmatvec(matvec, example_x)
+    square = True
+  except TypeError:
+    square = False
+    raise TypeError("The initialization `init` of solve_normal_cg is "
+                    "compulsory when `matvec` is nonsquare. It should "
+                    "have the same pytree structure as a solution. "
+                    "Typically, a pytree filled with zeros should work.")
+
```
froystig, line 170: It seems that the variable `square` is always `True` if we make it past this `try` block, even if we aren't in the square case. I think it still works out because of the previous comment about `_normal_matvec` being correct for non-square matvecs. Want to simplify or rework this function's logic? Is one way of doing `normal_matvec` preferred in the non-square case?

jaxopt/_src/linear_solve.py
```diff
@@ -109,14 +109,26 @@ def solve_cg(matvec: Callable,
...
+  rmatvec = _make_rmatvec(matvec, x)
+  return rmatvec(y)
+
+
+# FIXME: potentially rename this utility to _normal_square_matvec.
+# Not doing yet as this name is used in other files from the code base.
+def _normal_matvec(matvec, x=None):
+  """Computes A^T A x from matvec(x) = A x, where A is assumed square."""
   matvec_x, vjp = jax.vjp(matvec, x)
   return vjp(matvec_x)[0]
```
froystig, line 133: Two questions:

1. What is the intended behavior when `x` is `None`?
2. Assuming `x` is not `None`, doesn't this helper work for non-square `A` as well? It reads to me as equivalent to:
```python
  def _normal_matvec2(matvec, x):
    matvec_x = matvec(x)
    return jax.linear_transpose(matvec, x)(matvec_x)
```

jaxopt/_src/linear_solve.py
mblondel: Good catch, you're right `_normal_matvec` works in the nonsquare case too.

jaxopt/_src/linear_solve.py
mblondel: `x=None` was a leftover of something I had tried before. It was a mistake. Thanks for the catch!

jaxopt/_src/linear_solve.py
```diff
@@ -109,14 +109,26 @@ def solve_cg(matvec: Callable,
...
-  """Computes A^T x, from matvec(x) = A x, where A is square."""
+def _make_rmatvec(matvec, x):
   transpose = jax.linear_transpose(matvec, x)
-  return transpose(x)[0]
+  return lambda y: transpose(y)[0]
 
 
-def _normal_matvec(matvec, x):
-  """Computes A^T A x from matvec(x) = A x."""
+def _rmatvec(matvec, y, x=None):
```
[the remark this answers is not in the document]
mblondel: Done.

mblondel:
@froystig I addressed your comments and force-pushed.

froystig (approved):
Thanks, looks good!

PR #188, opened 2022-03-21
implement Levenberg-Marquardt first draft

amir-saadat:
Hi Mathieu, this is the first draft of Levenberg-Marquardt implementation. Please let me know your feedback. I also had some issues with the implicit differentiation unit test and it would be great if you can take a look and let me know if you encountered any issues. I am not passing on implicit_diff=True of test_double_exponential_fit_implicit_diff unit test.


mblondel (commented):
Thanks @amir-saadat! I made a pass on the code, still need to make a pass on the tests.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -0,0 +1,313 @@
...
+      ``residual_fun(x, *args, **kwargs)``.
+    maxiter: maximum number of iterations.
+    tol: tolerance.
+    implicit_diff: whether to enable implicit diff or autodiff of unrolled
+      iterations.
+    implicit_diff_solve: the linear system solver to use.
+    verbose: whether to print error on every iteration or not.
+      Warning: verbose=True will automatically disable jit.
+    jit: whether to JIT-compile the bisection loop (default: "auto").
+    unroll: whether to unroll the bisection loop (default: "auto").
```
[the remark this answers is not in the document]
amir-saadat: Good point thanks.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -0,0 +1,313 @@
...
+from jaxopt._src.linear_solve import solve_cholesky
+from jaxopt._src.tree_util import tree_l2_norm, tree_inf_norm, tree_sub, tree_add, tree_mul
+from jax.experimental import host_callback as hcb
+
+
+class LevenbergMarquardtState(NamedTuple):
+  """Named tuple containing state information."""
+  iter_num: int
+  damping_coeff: float
+  nu: float
```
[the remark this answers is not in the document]
amir-saadat: For sure, increase_factor works just fine.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -0,0 +1,313 @@
...
+    matvec = lambda v: self._jtj_op(init_params, v, *args, **kwargs)
+    gradient = self._jt_op(init_params, residual, *args, **kwargs)
+    nu = 2
+
+    grad_diff = tree_sub(tree_inf_norm(gradient), self.gtol)
+    grad_error = jax.lax.max(grad_diff, 0.0)
+
+    # found = jnp.linalg.norm(grad, jnp.inf) <= gtol
+    jtj_diag = self._jtj_diag_op(init_params, *args, **kwargs)
+    damping_coeff = self.damping_parameter * jnp.max(jtj_diag)
```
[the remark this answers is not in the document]
amir-saadat: For sure. Updated `damping_coeff` to `damping_factor`.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -0,0 +1,313 @@
...
+    delta_params = jnp.zeros_like(init_params)
+
+    return LevenbergMarquardtState(iter_num=jnp.asarray(0),
+                                   damping_coeff=damping_coeff,
+                                   nu=nu,
```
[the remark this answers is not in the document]
amir-saadat: Done.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -0,0 +1,313 @@
...
+    unroll: whether to unroll the bisection loop (default: "auto").
+  """
+  residual_fun: Callable
+
+  maxiter: int = 30
+  tol: float = 0.0
+
+  damping_parameter: float = 1e-6
+  xtol: float = 1e-8
+  gtol: float = 1e-8
```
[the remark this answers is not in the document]
amir-saadat: I am not sure if that would be possible, if we would like to keep the algorithm as is. The algorithm for this implementation suggests stopping either if gradients are small or the delta_params is small. If we have only one tol, we wouldn't be able to distinguish between the two, right? Does that address your question?

The drawback of the current usage is that tol needs to be zero in order to preserve the stopping criteria of the algorithm, at least this was the only way i could think of that doesn't require changing the `_cond_fun` for the `state.error` in the `base.py`.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -0,0 +1,313 @@
...
+      # to optimization and data fitting", L(0)-L(hlm)=0.5*hlm^T*(mu*hlm-g).
+      gain_ratio_denom = 0.5 * jnp.dot(
+          jnp.transpose(delta_params), (damping_coeff * delta_params - gradient))
+
+      # Current value of loss function F=0.5*||f||^2.
+      loss_next = 0.5 * jnp.dot(residual_next, residual_next)
+
+      gain_ratio = (loss_curr - loss_next) / gain_ratio_denom
+
+      def gain_ratio_test_true_func(cond_state):
```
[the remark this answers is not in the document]
amir-saadat: I think that is a great idea. I extracted two functions, the second one being called in the first one, and the first one is called within the `update` function. Please double check if consistent with your thoughts.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -0,0 +1,313 @@
...
+
+        damping_coeff = damping_coeff * jax.lax.max(1 / 3, 1 -
+                                                    (2 * gain_ratio - 1)**3)
+        nu = 2
+
+        return params, error, grad_error, damping_coeff, nu, gradient, residual
+
+      def gain_ratio_test_false_func(cond_state):
+
+        params, error, grad_error, damping_coeff, nu, _, _ = cond_state
```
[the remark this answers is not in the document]
amir-saadat: Done. Thanks for the suggestion.

jaxopt/_src/levenberg_marquardt.py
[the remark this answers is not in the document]
amir-saadat: Done. thanks.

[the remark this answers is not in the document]
amir-saadat: Done. thanks.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -0,0 +1,313 @@
...
+      delta_params = solve_cg(jtj_reg_matvec, state.gradient, init=state.delta)
+    elif self.solver == 'bicgstab':
+      delta_params = solve_bicgstab(
+            jtj_reg_matvec, state.gradient, init=state.delta)
+    elif self.solver == 'gmres':
+      delta_params = solve_gmres(
+            jtj_reg_matvec, state.gradient, init=state.delta)
+    elif self.solver == 'cholesky':
+      delta_params = solve_cholesky(jtj_reg_matvec, state.gradient)
+    elif self.solver == 'inv':
```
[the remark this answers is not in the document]
amir-saadat: Done. thanks for the suggestion.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -0,0 +1,313 @@
...
+      state: named tuple containing the solver state.
+    Returns:
+      (params, state)
+    """
+
+    # Current value of the loss function F=0.5*||f||^2.
+    loss_curr = state.loss
+
+    def jtj_reg_matvec(vec):
+      """The (J^T.J + damping_coeff*I) matrix vector product operator."""
```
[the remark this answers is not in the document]
amir-saadat: Done. thanks for the suggestion.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -0,0 +1,313 @@
...
+      return self._jtj_reg_op(params, state.damping_coeff, vec, *args, **kwargs)
+
+    if self.solver == 'cg':
```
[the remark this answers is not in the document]
amir-saadat: Done. Thanks.

tests/gauss_newton_test.py
```diff
@@ -83,7 +83,7 @@ def test_enzyme_reaction_parameter_fit(self):
                               onp.array([0.36183689, 0.55626653]))
 
 
-  @parameterized.product(implicit_diff=[True, False])
+  @parameterized.product(implicit_diff=[False, False])
```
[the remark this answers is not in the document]
amir-saadat: I haven't forgot this :). It is intentionally set as False as well, cause `True` doesn't pass the test. I basically tried what I did for Gauss-Newton, while the example here has many more data points (the number of x and y values are 40+). It would be great if you can take a look and see if I am using the implicit differentiation correctly. Thanks.

amir-saadat:
@mblondel thanks for your comments/suggestions. I think I have addressed most of your first round comments. I am still having issues with the implicit differentiation unit test. It would be great if you can take a look. Thank you

tests/gauss_newton_test.py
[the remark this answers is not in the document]
mblondel: But you didn't change anything in `gauss_newton.py` so why should the tests stop working?

jaxopt/_src/levenberg_marquardt.py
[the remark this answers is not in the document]
mblondel: So far, all solvers had a single stopping criterion, typically the l2 norm of  optimality_fun (the gradient in your case). Since optimality_fun is also used for implicit_diff, I think this makes sense and would be my preference for simplicity and consistency.

If you really want to keep , we would indeed need a way to override `_cond_fun` in base.py, as I'm not a fan of using `tol = 0`.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -0,0 +1,356 @@
...
+      **kwargs: additional keyword arguments to be passed to ``residual_fun``.
+
+    Returns:
+      state
+    """
+    # Compute actual values of state variables at init_param
+    residual, aux = self._fun_with_aux(init_params, *args, **kwargs)
+    gradient = self._jt_op(init_params, residual, *args, **kwargs)
+
+    grad_diff = tree_sub(tree_inf_norm(gradient), self.gtol)
```
mblondel, line 121: `tree_inf_norm` returns a scalar value so normally you can write

```suggestion
    grad_diff = tree_inf_norm(gradient) - self.gtol
```

You have a bunch of those in the code.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -0,0 +1,356 @@
...
+      delta_params = solve_inv(
+          matvec, state.gradient, ridge=state.damping_factor)
+    delta_params = -delta_params
+
+    # Define the functions required for the major conditional of the algorithm,
+    # which checks the magnitude of dparams and checks if it is small enough.
+    # for the value of dparams.
+
+    tree_mul_term = tree_mul(self.xtol,
+                             tree_add(tree_l2_norm(params), self.xtol))
```
mblondel, line 287: Same here and below. Normally `tree_l2_norm` returns a scalar so you don't need to use tree utils on top.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -0,0 +1,356 @@
...
+    Returns:
+      (params, state)
+    """
+
+    # Current value of the loss function F=0.5*||f||^2.
+    loss_curr = state.loss
+
+    def jtj_reg_matvec(vec):
+      """The (J^T.J + damping_factor*I) matrix vector product operator."""
+      return self._jtj_reg_op(params, state.damping_factor, vec, *args,
```
mblondel, line 257: Since you added the `ridge` option directly to the solvers I think this should be
```suggestion
      return self._jtj_op(params, state.damping_factor, vec, *args,
```
And in this case `def _jtj_reg_op` can be deleted.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -0,0 +1,356 @@
...
+    _, jvp_val = jax.jvp(fun_with_args, (params,), (vec,))
+    jtj_op_val, = vjpfun(jvp_val)
+    return jtj_op_val
+
+  def _jtj_reg_op(self, params, lam, vec, *args, **kwargs):
+    """Product of (J^T.J + lam*I) and vec, where J is jacobian of fun at params."""
+    return self._jtj_op(params, vec, *args, **
+                        kwargs) + lam * jnp.dot(jnp.eye(len(params)), vec)
+
+  def _jtj_diag_op(self, params, *args, **kwargs):
```
mblondel, line 352: Unfortunately, I think this is just as expensive as computing the full Jacobian once for all and store it. I see that this information is used for updating the damping factor. Does it mean that the Levenberg-Marquardt algorithm is not "matvec" friendly? Maybe @shoyer has some thoughts on this.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -0,0 +1,356 @@
...
+    def jtj_reg_matvec(vec):
+      """The (J^T.J + damping_factor*I) matrix vector product operator."""
+      return self._jtj_reg_op(params, state.damping_factor, vec, *args,
+                              **kwargs)
+
+    matvec = lambda v: self._jtj_op(params, v, *args, **kwargs)
+
+    if isinstance(self.solver, Callable):
+      delta_params = self.solver(
+          jtj_reg_matvec, state.gradient, init=state.delta)
```
mblondel, line 264: the `ridge` option is missing

tests/levenberg_marquardt_test.py
```diff
@@ -0,0 +1,149 @@
...
+        optimize_info.params - onp.array([1, 1]),
+        onp.array([-4.07e-09, -8.16e-09]),
+        atol=1e-11)
+    self.assertAllClose(optimize_info.state.iter_num, 16)
+
+  # This unit test is for checking rosenbrock function based on the example
+  # provided by K. Madsen & H. B. Nielsen in the book "Introduction to
+  # Optimization and Data Fitting" Example 6.10-. page 124. Data taken from
+  # immoptibox "A MATLAB TOOLBOX FOR OPTIMIZATION AND DATA FITTING" H.B. Nielsen
+  @parameterized.product(solver=['cg', 'bicgstab', 'gmres', 'cholesky', 'inv'])
```
mblondel, line 103: Maybe we should only accept 'cholesky', 'inv' and Callable. This way we don't have to test all possible solvers (we try to keep tests short when possible).

mblondel (changes requested):

tests/gauss_newton_test.py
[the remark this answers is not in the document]
amir-saadat: Oh my bad. I meant to avoid implicit differentiation in the LM file and I must have looked at the wrong file when making the change. Thank you for pointing that out.

jaxopt/_src/levenberg_marquardt.py
amir-saadat: This is one way to initialize the damping factor, but it is pretty typical. Although, this is only once during the optimization.

mblondel: If it's done only once in the initialization, then that's fine! Sorry I thought it was done in `update`.

jaxopt/_src/levenberg_marquardt.py
amir-saadat: Actually `jtj_reg_matvec ` can be deleted now, since all of the solvers are using `ridge` and I kept the `jtj_reg_matvec` for the custom solver. If we are assuming custom solver will have the `ridge` option, we can just remove this definition.

jaxopt/_src/levenberg_marquardt.py
amir-saadat: no worries, yeah it is just for `init_state`.

jaxopt/_src/levenberg_marquardt.py
amir-saadat: Done. thanks.

jaxopt/_src/levenberg_marquardt.py
amir-saadat: Done. thanks for catching.

jaxopt/_src/levenberg_marquardt.py
amir-saadat: Done. thanks for catching.

tests/levenberg_marquardt_test.py
amir-saadat: Sure :).

jaxopt/_src/levenberg_marquardt.py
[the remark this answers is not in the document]
amir-saadat: I chose to do something in between: I override the `_cond_fun` now but give the user the freedom to choose the `stop_criteria`. WDYT?

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -0,0 +1,372 @@
...
+
+  Attributes:
+    residual_fun: a smooth function of the form ``residual_fun(x, *args,
+      **kwargs)``.
+    maxiter: maximum increase_factormber of iterations.
+    damping_parameter: The parameter which adds a correction to the equation
+      derived for updating the coefficients using Gauss-Newton method. Please
+      see section 3.2. of K. Madsen et al. in the book "Methods for nonlinear
+      least squares problems" for more information.
+    stop_criteria: The criteria to use for the convergence of the while loop.
```
mblondel, line 78: I would call it `stopping_criterion` (`criteria` is plural)

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -0,0 +1,372 @@
...
+    algorithm 6.18 provided by K. Madsen & H. B. Nielsen in the book
+    "Introduction to Optimization and Data Fitting".
+  """
+  residual_fun: Callable
+
+  maxiter: int = 30
+  tol: float = 0.0
+
+  damping_parameter: float = 1e-6
+  stop_criteria: Literal['grad-l2', 'madsen-nielsen'] = 'madsen-nielsen'
```
mblondel, line 116: I think `grad-l2-norm` would be slightly more explicit.

Also I would use `grad-l2-norm` by default. This is consistent with all other solvers in JAXopt and uses `tol`.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -0,0 +1,372 @@
...
+  tol: float = 0.0
```
mblondel, line 117: If `grad-l2-norm` is the default, then `tol` should be non-zero.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -0,0 +1,372 @@
...
+  xtol: float = 1e-8
```
mblondel, line 118: `1e-8` may be very difficult to reach for float32 precision, which is the default precision we operate with. How about `1e-3`?

jaxopt/_src/levenberg_marquardt.py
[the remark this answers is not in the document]
mblondel: This works for me. In the future, we may want to provide a public API to do that. This implementation will be part of JAXopt so it's fine to override a private method but for external implementations, it's not ok.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -0,0 +1,372 @@
...
+    if isinstance(self.solver, Callable):
+      delta_params = self.solver(
+          matvec, state.gradient, ridge=state.damping_factor, init=state.delta)
+      if self.geodesic:
+        # jtrpp=JT * r" - J is jacobian and r" is second order dir. derivative
+        rpp = self._d2fvv_op(params, delta_params, delta_params, *args, **kwargs)
+        jtrpp = self._jt_op(params, rpp, *args, **kwargs)
+        delta_params += 0.5*self.solver(matvec, jtrpp, ridge=state.damping_factor)
+    elif self.solver == 'cg':
+      delta_params = solve_cg(
```
mblondel, line 289: Any reason to keep cg explicitly? Can't we support it via Callable? Especially since I see you don't test this branch anymore.

tests/levenberg_marquardt_test.py
```diff
@@ -0,0 +1,163 @@
...
+        onp.asarray([1, -1, 1, 2], dtype=float),
+        self.double_exponential_x,
+        self.double_exponential_y)
+
+    # Check the final result up to 4 sig digits.
+    self.assertArraysAllClose(
+        optimize_info.params, onp.array([4, -4, 4, 5], dtype=float), atol=1e-3)
+    # Convergence after 62 iterations.
+    if geodesic:
+      self.assertAllClose(optimize_info.state.iter_num, 20)
```
mblondel, line 138: Nice!

mblondel (changes requested):
Thanks a lot @amir-saadat for taking all my comments into account!

Could you also add `LevenvergMarquardt` to [api.rst](https://github.com/google/jaxopt/blob/main/docs/api.rst) and add a [changelog](https://github.com/google/jaxopt/blob/main/docs/api.rst) entry.

tests/levenberg_marquardt_test.py
amir-saadat: Yes! it is very exciting.

jaxopt/_src/levenberg_marquardt.py
amir-saadat: Done.

jaxopt/_src/levenberg_marquardt.py
amir-saadat: Done.

jaxopt/_src/levenberg_marquardt.py
amir-saadat: Done. Thanks.

amir-saadat:
Thanks @mblondel for great comments. I hope they are addressed now.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -0,0 +1,362 @@
...
+
+  maxiter: int = 30
+
+  damping_parameter: float = 1e-6
+  stop_criterion: Literal['grad-l2-norm', 'madsen-nielsen'] = 'grad-l2-norm'
+  tol: float = 1e-3
+  xtol: float = 1e-3
+  gtol: float = 1e-3
+
+  solver: Union[Literal['cholesky', 'inv'], Callable] = 'cholesky'
```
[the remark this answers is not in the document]
amir-saadat: Sure, I changed it to `solve_cg` I think that would make more sense since the matrix is symmetric PSD.

amir-saadat:
Thanks @mblondel for the comments. I updated api.rst, and changelog.rst. Please double check if I got it right. I added the changes for a new version v.3.02. The previous unit test was very slow (several min) and I decided to use another unit test that is similar to GaussNewton example. That is working fine now.

tests/levenberg_marquardt_test.py
```diff
@@ -0,0 +1,182 @@
...
+# limitations under the License.
+
+from absl.testing import absltest
+from absl.testing import parameterized
+from typing import Optional
+import unittest
+from unittest import SkipTest
+
+import jax
+from jax import test_util as jtu
```
mblondel, line 22: Remove

tests/levenberg_marquardt_test.py
```diff
@@ -0,0 +1,182 @@
...
+      return lm.run(
+          self.init_enzyme_reaction_coeffs,
+          substrate_conc,
+          self.rate_data).params
+    jac_custom = jax.jacrev(wrapper)(self.substrate_conc)
+
+    self.assertArraysAllClose(jac_num, jac_custom, atol=1e-2)
+
+if __name__ == '__main__':
+  absltest.main(testLoader=jtu.JaxTestLoader())
```
mblondel, line 182:
```suggestion
  absltest.main()
```

docs/changelog.rst
```diff
@@ -7,6 +7,17 @@ Main branch
 Bug fixes and enhancements
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+Version 0.3.2.
+--------------
```
mblondel, line 11: You can put your modification in "Main branch" for now. I am not sure what version number we should use yet.

amir-saadat: sounds good. thanks.

tests/levenberg_marquardt_test.py
amir-saadat: Applied. thanks.

tests/levenberg_marquardt_test.py
amir-saadat: Removed.

PR #195, opened 2022-04-04
increase x32 threshold in lm

amir-saadat:
@mblondel, this is to try the minor fix which can hopefully make the test more reliable.


mblondel (approved):
LGTM

PR #212, opened 2022-04-25
prettify adversarial training example

fabianp:
* Added plot of the difference between original and perturbed
  * Larger plots
  * A bit more of narrative text


mblondel (approved):
LGTM, sorry for the delay!

PR #216, opened 2022-05-02
better documentation for use_gamma keyword

fabianp:
Original one didn't explain what happens when use_gamma=False


mblondel (approved):
LGTM!

PR #218, opened 2022-05-02
LBFGS test on the Rosenbrock function

fabianp:
The Rosenbrock function is a 2-dimensional ill-conditioned function useful to test optimization algorithms.

This test is currently failing, but it shouldn't, as it's currently working on scipy (and jax through the ScipyMinimize wrapper.

Upon exit, the objective function is far from optimum but the step-size is 0, which makes me think that it's an issue in our line-search function


mblondel:
Thank you @fabianp. I suggest we merge it as a skipped test, to make it easier for me and others to debug.

fabianp:
Done, I marked it as skipped

Issue #207, opened 2022-04-19
Convenience API for QPs


mblondel:
Currently, the interface for QPs is like this:
```python
params_obj = (Q, c)
params_eq = (A, b)
params_ineq = (G, h)

osqp = jaxopt.OSQP()
osqp.run(init, params_obj, params_eq, params_ineq)
```
I propose that we also add the following convenience API:
```python
def guaranteed_quadratic_function(params, params_obj):
   [...]

osqp = jaxopt.OSQP(fun=guaranteed_quadratic_function)
osqp.run(init, params_obj, params_eq, params_ineq)
```
The idea is that we can, internally, automatically compute `Q` and `c` for the user using AD
```python
grad = jax.grad(self.fun)
c = grad(zeros, params_obj)
Q = jax.jacobian(grad)(zeros, params_obj)
```
Similarly, for solvers like OSQP that can leverage matvecs, we can derive the matvec of Q automatically.

CC @Algue-Rythme

mblondel:
or maybe
```python
def guaranteed_quadratic_function(params, *params_obj):
   [...]
```
instead

Algue-Rythme:
Good idea, I will take a look !

PR #228, opened 2022-06-06
Convenience API for OSQP

Algue-Rythme:
## What's new ?

Classes `OSQP` and `BoxOSQP` now support the (optional) parameter `fun`: a function promised to be a quadratic convex polynomial. Any function that can be written on the form `fun(x, params_Q)=0.5xQx+cx+cste` for some PSD matrix Q.

The advantages of matrix-vector products are leveraged to avoid the explicit computation of `Q` at any time.

Pros:
- easy writing of objective function for fast prototyping
- no need to disantangle quadratic `0.5xQx` and affine `cx+cste` terms: useful for least squares objectives: `||y-x||_2`

Cons:
- more calls to `jax.grad` are now required to retrieve `jnp.dot(c, x)` and to perform `Qx` matrix-vectors products
- this overhead _could_ be (partially) alleviated by an efficient XLA compiler with Common Sub Expression Elimination

## Discussion

First, I wanted to use the approach evocated in https://github.com/google/jaxopt/issues/207 but `Q = jax.jacobian(grad)(zeros, params_obj)` is troublesome:
1. Q is no longer guaranteed to be a 2D matrix: this causes `DenseLinearOperator` to fail
2. Q is not strictly necessary, and often too big for practical applications

Related to point 1. I am slowly realizing that generalizing this approach to other solvers, and to other linear operators, will require to support arbitrary tensors (i.e not only 2D matrices). It can be formalized with the notion of covariant indices (input dimension) and contravariant indices (output dimensions): see [this table on wikipedia](https://en.wikipedia.org/wiki/Tensor#Examples) so I add some precision about DenseLinearOperator in the doc: it currently only support 2D matrices.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,19 @@
 Changelog
 =========
 
+Development version
+___________________
+
+Bug fixes and enhancements
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- Add support for quadratic polynomial `fun` in :class:`jaxopt.BoxOSQP` and :class:`jaxopt.OSQP`.
+
+Contributors
+~~~~~~~~~~~~
+
+Louis Béthune.
+
 Version 0.4.2
 -------------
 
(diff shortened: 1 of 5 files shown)


Algue-Rythme:
Done !

mblondel:
Sorry, could you add the changelog entry under "Development version"? (we already released version 0.4.2)

mblondel:
Merged, thanks a lot @Algue-Rythme!

PR #233, opened 2022-06-09
Fix bug with `use_gamma=True`, add `stepsize` option, strengthen tests.

mblondel:
Thanks to @junpenglao for spotting.


jaxopt/_src/lbfgs.py
```diff
@@ -242,7 +247,7 @@ def update(self,
     (value, aux), grad = self._value_and_grad_with_aux(params, *args, **kwargs)
 
     start = state.iter_num % self.history_size
-    last = (start + self.history_size) % self.history_size
+    last = (start + self.history_size - 1) % self.history_size
```
junpenglao, line 250: rename to `previous` for clarity?

PR #236, opened 2022-06-09
Ensure the stepsize doesn't become too small.

mblondel:
The test with line search on the Rosenbrock function is now passing. I've added a mechanism so that if the stepsize becomes less than `min_stepsize`, then we restart it. Unfortunately, it has to be done in every solver (`LBFGS`, `NonlinearCG`, ...) separately because we don't have an API mechanism to do it in `BacktrackingLinesearch`. If the gradient norm becomes too small, it may be better to disable line search and switch to constant stepsize but I haven't implemented it yet, as this would make the branch data-dependent.

CC @fabianp @junpenglao @Algue-Rythme

### DIFF
--- a/jaxopt/_src/backtracking_linesearch.py
+++ b/jaxopt/_src/backtracking_linesearch.py
@@ -56,6 +56,7 @@ class BacktrackingLineSearch(base.IterativeLineSearch):
     c2: constant strictly less than 1 used by the (strong) Wolfe condition.
     decrease_factor: factor by which to decrease the stepsize during line search
       (default: 0.8).
+    max_stepsize: upper bound on stepsize.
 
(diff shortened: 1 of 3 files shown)


junpenglao (approved):

fabianp:
This is better than not doing anything although I'm concerned that this adds another hyperparameter max_stepsize. What do established solvers like scipy's LBFGS do in this scenario?

mblondel:
The line search in SciPy is using cubic interpolation, right? And it’s float64. So not entirely comparable. I plan to add the zoom linesearch currently in JAX core as a separate PR.

Can the stepsize become bigger than 1? Otherwise, this looks like a sane value to me.

Issue #243, opened 2022-06-17
Numerical issues in multiclass_logistic_loss


jvlmdr:
The current definition of multiclass_logistic_loss has a minor numerical issue:
```python
>>> jaxopt.loss.multiclass_logistic_loss(0, jnp.array([1e9, 1e9]))
DeviceArray(0., dtype=float32)  # expect log(2)
```

It is also not possible to use `-inf` to mask unused logits:
```python
>>> jaxopt.loss.multiclass_logistic_loss(0, jnp.array([0, 0, -jnp.inf]))
DeviceArray(nan, dtype=float32)  # expect log(2)
```

And, while less critical, it would be nice if it could handle positive infinity:
```python
>>> jaxopt.loss.multiclass_logistic_loss(0, jnp.array([jnp.inf, 0]))
DeviceArray(nan, dtype=float32)  # expect 0
```

One alternative implementation is:
```python
def multiclass_logistic_loss(label: int, logits: jnp.ndarray) -> float:
    return logsumexp((logits - logits[label]).at[label].set(0.))
```

mblondel:
Thanks a lot for reporting! We would welcome a fix and the corresponding unit tests.

mblondel:
Fixed in #245.

PR #245, opened 2022-06-19
Handle large numbers and -inf in losses

jvlmdr:
Adjusts order of operations to avoid precision issues (#243)

### DIFF
--- a/jaxopt/_src/loss.py
+++ b/jaxopt/_src/loss.py
@@ -57,7 +57,9 @@ def binary_logistic_loss(label: int, logit: float) -> float:
   # Softplus is the Fenchel conjugate of the Fermi-Dirac negentropy on [0, 1].
   # softplus = proba * logit - xlogx(proba) - xlogx(1 - proba),
   # where xlogx(proba) = proba * log(proba).
-  return softplus(logit) - label * logit
+  # Use -log sigmoid(logit) = softplus(-logit)
+  # and 1 - sigmoid(logit) = sigmoid(-logit).
+  return softplus(jnp.where(label, -logit, logit))
 
 
 # Multiclass classification.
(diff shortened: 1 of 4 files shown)


jaxopt/_src/loss.py
```diff
@@ -57,7 +57,9 @@ def binary_logistic_loss(label: int, logit: float) -> float:
   # Softplus is the Fenchel conjugate of the Fermi-Dirac negentropy on [0, 1].
   # softplus = proba * logit - xlogx(proba) - xlogx(1 - proba),
   # where xlogx(proba) = proba * log(proba).
-  return softplus(logit) - label * logit
```
mblondel, line 60: Could you add this implementation to the tests and check that the new one recovers the same value?

jaxopt/_src/loss.py
```diff
@@ -93,10 +96,9 @@ def multiclass_sparsemax_loss(label: int, scores: jnp.ndarray) -> float:
...
-  n_classes = scores.shape[0]
-  one_hot = jax.nn.one_hot(label, n_classes)
   proba = projection_simplex(scores)
   # Fenchel conjugate of the Gini negentropy, defined by:
   # cumulant = jnp.dot(proba, scores) + 0.5 * jnp.dot(proba, (1 - proba)).
-  cumulant = 0.5 + jnp.dot(proba , scores - 0.5 * proba)
-  return cumulant - jnp.dot(scores, one_hot)
+  scores = (scores - scores[label]).at[label].set(0.0)
+  return (jnp.dot(proba, jnp.where(proba, scores, 0.0))
+          + 0.5 * (1.0 - jnp.dot(proba, proba)))
```
mblondel, line 104: Same here, could you check in the tests against the old implementation.

mblondel (commented):
Thanks for tackling all these edge cases. This is great!

Before we merge, could you add tests to check against the old implementation of the logistic and sparsemax losses? The old implementations are less robust but more readable so this is a great want to check correctness. The current tests are just sanity checks but they don't really ensure that the implementation is correct.

jvlmdr:
Thanks for reviewing this! I have added those tests.

I thought I would mention that, since this implementation uses `ndarray.at`, it will only work when called on `jnp.ndarray` and not `onp.ndarray` types.

I could remove this, I just thought it was nicer for the purpose of gradient computation to force the difference to be zero, although it should be fine in practice. (It also enables the loss functions to support `+inf` for the positive label (avoiding `inf - inf => nan`), although this doesn't seem like an important use case. I didn't add a test for this because it would have complicated the implementation of the simplex projection to make the `sparsemax` loss pass.)

mblondel:
Can't you just wrap the array with `logits = jnp.array(logits)` at the beginning?

jvlmdr:
Yep, done. I left the typing annotation as `logits: jnp.ndarray`

mblondel:
This is ready for merge thanks a lot! Could you squash commits? Thanks in advance!

jvlmdr:
Yep, done! Thank you!

Issue #254, opened 2022-06-22
Error when running Multiclass linear SVM (without intercept) example


QB3:
When running the Multiclass linear SVM (without intercept) example, (https://jaxopt.github.io/stable/auto_examples/constrained/multiclass_linear_svm.html) the QP solver yields the following error:

`OSQP solution solution:
Traceback (most recent call last):
  File "multiclass_linear_svm.py", line 144, in <module>
    app.run(main)
  File "/home/qbe/anaconda3/lib/python3.7/site-packages/absl/app.py", line 303, in run
    _run_main(main, args)
  File "/home/qbe/anaconda3/lib/python3.7/site-packages/absl/app.py", line 251, in _run_main
    sys.exit(main(argv))
  File "multiclass_linear_svm.py", line 127, in main
    W_osqp = multiclass_linear_svm_osqp(X, Y, l2reg)
  File "multiclass_linear_svm.py", line 105, in multiclass_linear_svm_osqp
    init_params = osqp._box_osqp.init_params(init_x=None, **hyper_params)
  File "/home/qbe/jaxopt/jaxopt/_src/osqp.py", line 410, in init_params
    init_z = projection_box(self.matvec_A(params_eq)(init_x), params_ineq)
  File "/home/qbe/jaxopt/jaxopt/_src/osqp.py", line 47, in projection_box
    return tree_map(_clip_safe, x, lower, upper)
  File "/home/qbe/jax/jax/_src/tree_util.py", line 180, in tree_map
    all_leaves = [leaves] + [treedef.flatten_up_to(r) for r in rest]
  File "/home/qbe/jax/jax/_src/tree_util.py", line 180, in <listcomp>
    all_leaves = [leaves] + [treedef.flatten_up_to(r) for r in rest]
ValueError: Expected list, got None.`

I guess the API of `OSQP` changed, but I did not managed to find a quick fix.

I am using the `main` branch of the developement version of  `jaxopt`

mblondel:
Confirmed on my side. CC @Algue-Rythme

PR #256, opened 2022-06-24
fix init_params bug in multiclass_svm example

Algue-Rythme:
Fix https://github.com/google/jaxopt/issues/254

### DIFF
--- a/examples/constrained/multiclass_linear_svm.py
+++ b/examples/constrained/multiclass_linear_svm.py
@@ -102,7 +102,6 @@ def multiclass_linear_svm_osqp(X, Y, l2reg):
                       params_eq=(None, b),
                       params_ineq=(None, h))
   
-  init_params = osqp._box_osqp.init_params(init_x=None, **hyper_params)
   sol, _ = osqp.run(init_params=None, **hyper_params)
   return sol.primal
 
--- a/jaxopt/_src/osqp.py
+++ b/jaxopt/_src/osqp.py
@@ -962,6 +962,8 @@ class OSQP(base.Solver):
       (params, state), ``params = (primal_var, dual_var_eq, dual_var_ineq)``
     """
     assert params_obj is not None
+    if init_params is None:
+      init_params = self.init_params(None, params_obj, params_eq, params_ineq)
     init_params, hyper_params, eq_ineq_size = OSQP_to_BoxOSQP.transform(self.matvec_A_box,
                                                                         init_params, params_obj,
                                                                         params_eq, params_ineq)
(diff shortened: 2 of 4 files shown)


mblondel:
Thanks @Algue-Rythme! The CI failures are unrelated, I'm investigating what's going on.

mblondel:
The failures happen with numpy 1.23.0 https://github.com/google/jax/issues/11238

mblondel:
Please change the numpy line to `numpy<1.23.0` in `requirements.txt`. This should allow the tests to pass. We will change it back later when the upstream issue is fixed.

examples/constrained/multiclass_linear_svm.py
```diff
@@ -102,8 +102,8 @@ def matvec_G(_, beta):
                       params_eq=(None, b),
                       params_ineq=(None, h))
   
-  init_params = osqp._box_osqp.init_params(init_x=None, **hyper_params)
-  sol, _ = osqp.run(init_params=None, **hyper_params)
+  init_params = osqp.init_params(init_x=None, **hyper_params)
+  sol, _ = osqp.run(init_params=init_params, **hyper_params)
```
mblondel, line 106: Since you take care of the `init_params is None` case in `osqp.py`, can't you just do 
```python
sol, _ = osqp.run(**hyper_params)
```

examples/constrained/multiclass_linear_svm.py
```diff
@@ -102,8 +102,8 @@ def matvec_G(_, beta):
                       params_eq=(None, b),
                       params_ineq=(None, h))
   
-  init_params = osqp._box_osqp.init_params(init_x=None, **hyper_params)
-  sol, _ = osqp.run(init_params=None, **hyper_params)
+  init_params = osqp.init_params(init_x=None, **hyper_params)
```
mblondel, line 105: Unrelated to this example, but why is it called `init_x`? Shouldn't this be called `init_params` instead?

examples/constrained/multiclass_linear_svm.py
mblondel: The issue with the `init_params()` method is that it's not really part of our standard API so it will be surprising for some users.

Algue-Rythme:
OSQP now fail to detect infeasability in projection polyhedron: I am on it.

mblondel (approved):
LGTM, thanks @Algue-Rythme

examples/constrained/multiclass_linear_svm.py
Algue-Rythme: That would make sense indeed.

Issue #169, opened 2022-02-03
Complex gradients


GeoffNN:
Opening this as a nota bene. When optimizing over complex parameters, the gradient must be conjugated. If any optimizer relies on a second order moment (eg Adam), it must also use the complex module squared instead of just the parameter squared. Current jaxopt solvers might be affected as well.

I realized this while using Optax on a model with complex weights; thought it might be good to incorporate this in jaxopt solvers as well, as users might 1) not be aware of this and 2) it's really hard to debug on the user side.

GeoffNN:
For instance, tree_l2_norm would currently give incorrect results on complex parameters.

https://github.com/google/jaxopt/blob/eb6e75dfee1d25cc2b206ad1410668c576bf6750/jaxopt/_src/tree_util.py#L84

mblondel:
+1 on fixing this, thanks for catching

GeoffNN:
Hey! Sorry, I was interning this summer and off of github. I'll start checking this out!

PR #260, opened 2022-06-28
Native pytree support in zoom line search.

mblondel:


jaxopt/_src/zoom_linesearch.py
```diff
@@ -272,21 +276,24 @@ def zoom_linesearch(f, xk, pk, old_fval=None, old_old_fval=None, gfk=None, c1=1e
...
-    # FIXME: directly acept a value_and_grad function to avoid recompilations.
-    phi, g = jax.value_and_grad(f)(xk + t * pk)
-    dphi = jnp.real(_dot(g, pk))
+    #t = jnp.array(t, dtype=pk.dtype)
+
+    xkp1 = tree_add_scalar_mul(xk, t, pk)
+
+    # FIXME: directly accept a value_and_grad function to avoid recompilations.
+    phi, g = jax.value_and_grad(f)(xkp1)
+    dphi = tree_real(tree_vdot(g, pk))
```
junpenglao, line 289: How often do we have this pattern? I am not sure how expensive is the additional tree map, maybe it is better to a single function that we map to the input?
```python
real_vdot = lambda x, y: jnp.real(_vdot_safe(x, y))
tree_real_vdot = lambda x, y: tree_map(real_vdot, tree)
```

mblondel: Good question. I wonder if JAX's `jit` is able to optimize this automatically. If not, happy to add the utility you suggest. BTW, we don't have good support for complex numbers in JAXopt, see #169.

mblondel: Hmm, I actually just realized that `tree_real` isn't necessary, just `jnp.real` will be enough, since `tree_vdot` returns a scalar.

PR #268, opened 2022-07-07
LBFGS benchmark using a CNN.

mblondel:


benchmarks/lbfgs_benchmark.py
```diff
@@ -108,6 +114,91 @@ def run_multiclass_logreg():
...
+      x = nn.avg_pool(x, window_shape=(2, 2), strides=(2, 2))
+      x = x.reshape((x.shape[0], -1))  # flatten
+      x = nn.Dense(features=self.net_width*4)(x)
+      x = nn.relu(x)
+      x = nn.Dense(features=self.num_classes)(x)
+      return x
+
+  # Hide any GPUs from TensorFlow. Otherwise TF might reserve memory and make
+  # it unavailable to JAX.
+  tf.config.experimental.set_visible_devices([], 'GPU')
```
junpenglao, line 149: This is interesting - why do we have TF here at all? Is it because we import tfds above?

fabianp: I believe this is due to tensorflow datasets, which imports (parts of) tensorflow. Without this line, tf occupies the memory of the GPU and jax can't use it

PR #269, opened 2022-07-08
Allow formulas to be well displayed in html pages

amelieH:
nbsphinx seems to be the right extension to display formulas from notebooks.
I removed the source_suffix list because of https://github.com/spatialaudio/nbsphinx/issues/310. So we shouldn't put ipynb in the extension list (to use nbsphinx), but if we just keep md and rst, the whole notebook appears raw.

I also had to remove the myst-nb extension that seemed in conflict with nbsphinx.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -122,6 +122,7 @@ html_css_files = [
 nb_execution_mode = "force"
 nb_execution_allow_errors = False
 nb_execution_fail_on_error = True  # Requires https://github.com/executablebooks/MyST-NB/pull/296
+myst_enable_extensions = ['dollarmath'] # To display maths in notebook
 
 # Notebook cell execution timeout; defaults to 30.
 nb_execution_timeout = 100


mblondel:
Thanks a lot for the investigation! I think your solution should be fine but I would just like to ensure that we don't lose any important feature by switching from `myst-nb` to `nbsphinx`. According to https://docs.readthedocs.io/en/stable/guides/jupyter.html ("Existing relevant extensions"), it seems like we don't but maybe @fabianp or @jakevdp can confirm. 

Some options in `conf.py` like `nb_execution_excludepatterns` (which is important to exclude long-running notebooks) may be specific to `myst-nb` but I'm not sure.

It seems like in the JAX documentation, the maths render correctly despite using myst-nb (e.g., https://jax.readthedocs.io/en/latest/notebooks/Custom_derivative_rules_for_Python_code.html) so we might miss something in the configuration.

amelieH:
You are right, after looking more at the JAX documentation conf.py file, I saw that they just add
myst_enable_extensions = ['dollarmath']
to the myst-nb options.

That works for us too.

This solution is much simpler and shouldn't have side effects.

Issue #116, opened 2021-12-06
Avoid state.aux=None in state returned by initial_state


mblondel:
Currently, when `has_aux=True`, `state.aux` is `None` when `state` is returned by `init_state` and `state.aux` is equal to `fun(params, *args, **kwargs)[1]` when `state` is returned by `update`. This is problematic as it can trigger a jit recompilation. One way would be to set `state.aux` to some dummy values of the correct type when returned by `init_state`.

Algue-Rythme:
Is there a better way than calling `self.fun_with_aux` using parameters of `init_state` just to retrieve the `aux` value ? That would do the trick; the only drawback is to pay the price of calling `fun_with_aux` "for nothing".

mblondel:
One idea would be to trace through `fun` in order to retrieve the return types but not sure if it's feasible.

mblondel:
https://jax.readthedocs.io/en/latest/jax.html#jax.eval_shape (thanks to @josipd for the tip)

PR #273, opened 2022-07-12
Ensure aux type consistency in init_state and update.

### LINKED ISSUE #258: Issue with `scan` + OptaxSolver(., ., has_aux=True)
I believe the following lines of code could be tweaked so that we can use `OptaxSolver(., ., has_aux=True)` inside an update funcation that we pass to `jax.lax.scan` (an example of this type of training loop is shown below). It appears that the line `aux=None` is the problem as "scan carry output and input must have same type structure"

https://github.com/google/jaxopt/blob/ea89be7b42fe78eb1abfefd74bd872429eeda39b/jaxopt/_src/optax_wrapper.py#L94-L98

```
 @partial(jax.jit, static_argnums=(2,))
  def train(theta, state, epochs):

      def update(carry, t):
          theta, state = carry 
          theta, state = solver.update(params=theta, state=state, init_inner= jnp.zeros(X.shape[1]), data=data)
          return (theta, state), (state.value, state.aux)
      (theta_opt,_), (val_losses, train_losses) = jax.lax.scan(update, (theta, state), None, length=epochs)
      return val_losses, train_losses
```

mblondel:
Fixes #258 and #116

### DIFF
--- a/jaxopt/_src/armijo_sgd.py
+++ b/jaxopt/_src/armijo_sgd.py
@@ -214,17 +214,20 @@ class ArmijoSGD(base.StochasticSolver):
     Returns:
       state
     """
-    del args, kwargs  # Not used.
-
     if self.momentum == 0:
       velocity = None
     else:
       velocity = tree_zeros_like(init_params)
 
+    if self.has_aux:
+      _, aux = self.fun(init_params, *args, **kwargs)
+    else:
+      aux = None
+
     return ArmijoState(iter_num=jnp.asarray(0),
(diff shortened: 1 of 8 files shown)


fllinares (approved):
LGTM, thanks Mathieu!

I just have one question / note for the future. I wonder if there are cases for which `self.fun` should be jitted. Though I guess if that was a problem (e.g. OOM when evaluating `self.fun` without jit) the user could jit `init_state` instead.

PR #274, opened 2022-07-14
Sparse root VJP for Lasso penalty

tristandeleu:
We have been working with @QB3 on a rework of #17, adding support functions to enable implicit differentiation with sparsity-inducing penalties. The main difference is that we are masking the linear system to solve in `root_vjp` based on the support of the solution, instead of restricting it to its support, in order to be jit-compatible. For now, this functionality has only been added to `ProximalGradient`, but we can add it to other solvers as well if we agree on the API.

Specifying the support explicitly allows us to ensure that the Jacobian will be non-zero only for coordinates in the support of the solution. In the case of Lasso, this also allows us to use CG to solve the linear system, instead of Normal-CG, since we ensure that the matrix to invert is symmetric. We have written a benchmark to showcase the advantages of masking the linear system to the support only in Lasso:
```
Number of samples: 100
Number of features: 1000
Size of the support of the solution: 4
Numerical Jacobian: [-0.64471806 -0.72820061 -0.87170093 -0.76674667]
Jacobian w/o support, CG (14.901 sec.): [ 1.6838026e+09 -2.0496195e+09  1.6045356e+09  1.1004994e+08] (size of the support: 4)
Jacobian w/o support, normal CG (2.114 sec.): [-0.644696   -0.7282007  -0.8717052  -0.76675165] (size of the support: 1000)
Jacobian w/ restricted support (1.404 sec.): [-0.6446959  -0.7282008  -0.8717053  -0.76675177] (size of the support: 4)
Jacobian w/ masked support (1.482 sec.): [-0.64469576 -0.72820073 -0.87170535 -0.76675177] (size of the support: 4)
```

(more details about the PR to be added)

### DIFF
--- a/benchmarks/sparse_root_vjp_benchmark.py
+++ b/benchmarks/sparse_root_vjp_benchmark.py
@@ -56,9 +56,12 @@ def benchmark_vjp(vjp, X, supp=None, desc=''):
   result.block_until_ready()
   delta_t = time.time() - t0
 
+  size_support = onp.sum(result != 0)
+
   if supp is not None:
     result = result[supp]
-  print(f'{desc} ({delta_t:.3f} sec.): {result}')
+  print(f'{desc} ({delta_t:.3f} sec.): {result} '
+        f'(size of the support: {size_support:d})')
 
 
 def main(argv: Sequence[str]) -> None:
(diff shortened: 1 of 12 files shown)


mblondel:
Thanks @tristandeleu and @QB3 for this work, this is very interesting! If I understand your main point correctly, masking is almost as fast as restricting but is jit friendly and allows to use CG in the case of the lasso? 

I would be curious to know the benchmark results on GPU.

By the way, I see the PR is currently marked as draft. Let me know when you would like me to review :)

PR #278, opened 2022-08-01
Fix: parse tol to scipy minimize.

pipme:
`tol` is not parsed to `scipy.optimize.minimize` in `ScipyMinimize`.


mblondel:
Thanks for spotting and for the contribution! It seems that the test `test_logreg_with_intercept` is now failing. Can you fix it and force-push? Thanks!

pipme:
Hi, I changed the `tol` to `1e-5` and now the test is passed.

Issue #272, opened 2022-07-12
ScipyMinimize doesn't expose a maxiter option


mblondel:
Currently, `ScipyMinimize` doesn't expose a `maxiter` option like all other solvers in JAXopt. Instead, it's part of the `options` dict.

mblondel:
#284 partial fixes this issue but it still needs to be done for other wrappers.

PR #280, opened 2022-08-08
Typo: ScipyMinimize doesn't have a maxiter argument

fabianp:


mblondel (approved):
LGTM, though it would be great if we can solve #272 too.

PR #289, opened 2022-08-24
Add callback mechanism to ScipyMinimize

fabianp:
This is an option that exists in scipy.optimize.minimize, but which we failed to expose, and its essential for detailed benchmarks

### DIFF
--- a/jaxopt/_src/scipy_wrappers.py
+++ b/jaxopt/_src/scipy_wrappers.py
@@ -17,7 +17,7 @@
 # TODO(fllinares): add support for `LinearConstraint`s.
 # TODO(fllinares): add support for methods requiring Hessian / Hessian prods.
 # TODO(fllinares): possibly hardcode `dtype` attribute, as likely useless.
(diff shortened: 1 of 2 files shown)


fabianp:
wait, I forgot to add the docstring for this option!

fabianp:
hmmm, one thing that I had underestimated is the flattening / unflattening of pytrees. A flattened numpy array is what gets passed to the callback function, even when the input to ScipyMinimize is a pytree. 

Should we be doing the work on unflattening the numpy array so that the user codes the callback expecting a pytree? This seems the most reasonable to me. What do you think @mblondel ?

mblondel:
I agree a pytree would be more natural.

fabianp:
I documented the callback option in the docstring and added the unraveling for the callback (and a test for it). I think this is now ready for review

mblondel:
Thanks @fabianp! 

@fllinares can you take a look when you get some time? Thanks

fllinares (approved):
Thanks a lot Fabian, LGTM!!

PR #295, opened 2022-08-26
Faster tests? [Don't merge]

fabianp:
Don't merge, I want to use this PR to see how much time we would be gaining

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -3,3 +3,4 @@ optax>=0.0.9
 pytest-xdist
 scikit-learn>=0.24.1
 cvxopt
+pytest-xdist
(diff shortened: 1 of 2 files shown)


requirements_test.txt
```diff
@@ -3,3 +3,4 @@ optax>=0.0.9
 pytest-xdist
 scikit-learn>=0.24.1
 cvxopt
+pytest-xdist
```
mblondel, line 6: Already on line 3 ;-)

fabianp: LOL true

fabianp:
runtime for tests seems to be divided by two, but there's a random failure in 3.10

PR #296, opened 2022-08-27
Use a tuple for marginals.

mblondel:


mblondel:
@liutianlin0121 As we discussed, I made the following changes.

liutianlin0121:
> @liutianlin0121 As we discussed, I made the following changes.

Thanks!

PR #297, opened 2022-08-27
Prepare release 0.5.

mblondel:


fabianp (approved):
minor nit: I would use 0.5.0 for consistency with future versions

thanks for doing this!

mblondel:
I'm inclined to keep it as is because this is the format I've been using so far, see https://github.com/google/jaxopt/releases

PR #304, opened 2022-09-07
Remove author information from notebooks

fabianp:


mblondel:
We should probably indicate the license.

fabianp:
good point

fabianp:
superceded by #336

PR #310, opened 2022-09-23
remove $ from the prompt

fabianp:
otherwise when you copy it using the copy button it results in an invalid command


mblondel (approved):
LGTM

PR #315, opened 2022-10-03
FIX: wrong link for open in colab

fabianp:


mblondel:
There seems to be unrelated failures in the Levenberg–Marquardt module...

mblondel (approved):

PR #316, opened 2022-10-03
Implement value_and_grad option.

mblondel:


jaxopt/_src/gradient_descent.py
```diff
@@ -33,19 +33,28 @@ class GradientDescent(ProximalGradient):
     fun: a smooth function of the form ``fun(parameters, *args, **kwargs)``,
       where ``parameters`` are the model parameters w.r.t. which we minimize
       the function and the rest are fixed auxiliary parameters.
+    value_and_grad: whether ``fun`` just returns the value (False) or both
+      the value and gradient (True).
+    has_aux: whether ``fun`` outputs auxiliary data or not.
```
fllinares, line 38: Probably obvious from the context, but the documentation for `has_aux` only describes the output signature of `fun` when `has_aux == True`. Should we explain `has_aux == False` too?

fllinares (approved):
Thanks a lot Mathieu, looks great! Just left a minor question regarding the doc (which applies to all solvers).

froystig (approved):
Looks good!

mblondel:
Thanks both for the reviews! I addressed Felipe's comment.

PR #323, opened 2022-10-07
Adding the possibility to stop LBFGS after a line search failure

zaccharieramzi:
This PR addresses the issue mentioned [here](https://github.com/google/jaxopt/discussions/322#discussioncomment-3824631), by allowing the user to decide if they want to have LBFGS fail when the line search fails or not using a `allow_line_search_fail` attribute (set to `False` by default to ensute backward compatibility).

I am not sure exactly which unit tests I should add for this new addition : typically it will fail on the Rosenbrock function.

### DIFF
--- a/jaxopt/_src/lbfgs.py
+++ b/jaxopt/_src/lbfgs.py
@@ -144,6 +144,7 @@ class LbfgsState(NamedTuple):
   rho_history: jnp.ndarray
   gamma: jnp.ndarray
   aux: Optional[Any] = None
+  failed_linesearch: bool = False
 
 
 @dataclass(eq=False)
@@ -173,6 +174,8 @@ class LBFGS(base.IterativeSolver):
       or a callable specifying the **positive** stepsize to use at each iteration.
     linesearch: the type of line search to use: "backtracking" for backtracking
       line search or "zoom" for zoom line search.
+    stop_if_linesearch_fails: whether to stop iterations if the line search fails.
+      When True, this matches the behavior of core JAX.
     maxls: maximum number of iterations to use in the line search.
     decrease_factor: factor by which to decrease the stepsize during line search
       (default: 0.8).
@@ -212,6 +215,7 @@ class LBFGS(base.IterativeSolver):
 
   stepsize: Union[float, Callable] = 0.0
   linesearch: str = "zoom"
+  stop_if_linesearch_fails: bool = False
   condition: str = "strong-wolfe"
   maxls: int = 15
   decrease_factor: float = 0.8
@@ -230,6 +234,14 @@ class LBFGS(base.IterativeSolver):
   unroll: base.AutoOrBoolean = "auto"
 
   verbose: bool = False
+
+  def _cond_fun(self, inputs):
+    _, state = inputs[0]
+    if self.verbose:
+      print("error:", state.error)
+    # We continue the optimization loop while the error tolerance is not met and,
+    # either failed linesearch is disallowed or linesearch hasn't failed.
+    return (state.error > self.tol) & (~self.stop_if_linesearch_fails | ~state.failed_linesearch)
 
   def init_state(self,
                  init_params: Any,
@@ -257,7 +269,8 @@ class LBFGS(base.IterativeSolver):
                       y_history=init_history(init_params, self.history_size),
                       rho_history=jnp.zeros(self.history_size, dtype=dtype),
                       gamma=jnp.asarray(1.0, dtype=dtype),
-                      aux=aux)
+                      aux=aux,
+                      failed_linesearch=jnp.asarray(False))
 
   def update(self,
              params: Any,
@@ -284,7 +297,8 @@ class LBFGS(base.IterativeSolver):
                                   start=start)
     descent_direction = tree_scalar_mul(-1.0, product)
 
-    if not isinstance(self.stepsize, Callable) and self.stepsize <= 0:
+    use_linesearch = not isinstance(self.stepsize, Callable) and self.stepsize <= 0
+    if use_linesearch:
       # with line search
 
       if self.linesearch == "backtracking":
@@ -352,6 +366,11 @@ class LBFGS(base.IterativeSolver):
     else:
       gamma = jnp.array(1.0)
 
+    if use_linesearch and self.linesearch == "zoom":
+      failed_linesearch = ls_state.failed
+    else:  # backtracking linesearch doesn't support failed state yet
+      failed_linesearch = jnp.asarray(False)
+
     new_state = LbfgsState(iter_num=state.iter_num + 1,
                            value=new_value,
                            stepsize=jnp.asarray(new_stepsize),
@@ -363,7 +382,8 @@ class LBFGS(base.IterativeSolver):
                            # FIXME: we should return new_aux here but
                            # BacktrackingLineSearch currently doesn't support
                            # an aux.
-                           aux=aux)
+                           aux=aux,
+                           failed_linesearch=failed_linesearch)
 
     return base.OptStep(params=new_params, state=new_state)
 


jaxopt/_src/lbfgs.py
```diff
@@ -142,7 +142,9 @@ class LbfgsState(NamedTuple):
   s_history: Any
   y_history: Any
   rho_history: jnp.ndarray
+  gamma: jnp.ndarray
   aux: Optional[Any] = None
+  failed: bool = False
```
mblondel, line 147:
```suggestion
  failed_linesearch: bool = False
```

jaxopt/_src/lbfgs.py
```diff
@@ -211,6 +213,7 @@ class LBFGS(base.IterativeSolver):
 
   stepsize: Union[float, Callable] = 0.0
   linesearch: str = "zoom"
+  allow_line_search_fail: bool = False
```
mblondel, line 216:
```suggestion
  allow_failed_linesearch: bool = False
```

jaxopt/_src/lbfgs.py
```diff
@@ -230,6 +233,12 @@ class LBFGS(base.IterativeSolver):
 
   verbose: bool = False
 
+  def _cond_fun(self, inputs):
+    _, state = inputs[0]
+    if self.verbose:
+      print("error:", state.error)
+    return (state.error > self.tol) & ~state.failed
```
mblondel, line 240:
```suggestion
    return (state.error > self.tol) & (self.allow_failed_linesearch | ~state.failed)
```

jaxopt/_src/lbfgs.py
```diff
@@ -255,7 +264,9 @@ def init_state(self,
                       s_history=init_history(init_params, self.history_size),
                       y_history=init_history(init_params, self.history_size),
                       rho_history=jnp.zeros(self.history_size, dtype=dtype),
-                      aux=aux)
+                      gamma=jnp.asarray(1.0, dtype=dtype),
+                      aux=aux,
+                      failed=jnp.asarray(False))
```
mblondel, line 269: I'm not sure we need `asarray` for booleans. CC @froystig

jaxopt/_src/lbfgs.py
```diff
@@ -351,17 +356,29 @@ def update(self,
     y_history = update_history(state.y_history, y, last)
     rho_history = update_history(state.rho_history, rho, last)
 
+    if self.use_gamma:
+      gamma = compute_gamma(s_history, y_history, last)
+    else:
+      gamma = jnp.array(1.0)
+
+    if self.linesearch == "zoom" and self.allow_line_search_fail:
```
mblondel, line 364:
```suggestion
    if self.linesearch == "zoom":
```

jaxopt/_src/lbfgs.py
```diff
@@ -351,17 +356,29 @@ def update(self,
...
+      failed = ls_state.failed
+    else:
```
mblondel, line 366:
```suggestion
    else:  # backtracking linesearch doesn't support failed state yet
```

mblondel (commented):
Thanks a lot for doing this. I left some comments and suggestions.

jaxopt/_src/lbfgs.py
zaccharieramzi:
```suggestion
    return (state.error > self.tol) & (~self.allow_failed_linesearch | ~state.failed_linesearch)
```

I think there was a logic error here, will check it out

jaxopt/_src/lbfgs.py
zaccharieramzi: I can try without and tell you

zaccharieramzi: it works without the `asarray`, so removed it

jaxopt/_src/lbfgs.py
zaccharieramzi: There was indeed a logic error so my suggestion was the one I picked

zaccharieramzi:
@mblondel went over your comments.
Only thing is that like I mentioned, I don't have a unit test for this.

jaxopt/_src/lbfgs.py
```diff
@@ -230,6 +233,12 @@ class LBFGS(base.IterativeSolver):
 
   verbose: bool = False
 
+  def _cond_fun(self, inputs):
+    _, state = inputs[0]
+    if self.verbose:
+      print("error:", state.error)
+    return (state.error > self.tol) & (~self.allow_failed_linesearch | ~state.failed_linesearch)
```
mblondel, line 240:
```suggestion
    # We continue the optimization loop while the error tolerance is not met and,
    # either failed linesearch is disallowed or linesearch hasn't failed.
    return (state.error > self.tol) & (~self.allow_failed_linesearch | ~state.failed_linesearch)
```

jaxopt/_src/lbfgs.py
```diff
@@ -255,7 +264,9 @@ def init_state(self,
                       s_history=init_history(init_params, self.history_size),
                       y_history=init_history(init_params, self.history_size),
                       rho_history=jnp.zeros(self.history_size, dtype=dtype),
-                      aux=aux)
+                      gamma=jnp.asarray(1.0, dtype=dtype),
+                      aux=aux,
+                      failed_linesearch=jnp.asarray(False))
```
mblondel, line 269: Without `asarray` I guess?

mblondel (commented):
Thanks. LGTM apart from 2 minor comments.

jaxopt/_src/lbfgs.py
zaccharieramzi: yup changing

jaxopt/_src/lbfgs.py
```diff
@@ -211,6 +213,7 @@ class LBFGS(base.IterativeSolver):
 
   stepsize: Union[float, Callable] = 0.0
   linesearch: str = "zoom"
+  allow_failed_linesearch: bool = False
```
mblondel, line 216: So the default behavior now is the same as that of LBFGS in JAX core, right?  (just to confirm)

jaxopt/_src/lbfgs.py
mblondel: Thanks for checking. I see that `bool(jnp.asarray(True))` gives `True` and similarly for `False` so all is good.

mblondel (approved):
LGTM apart from squashing commits. Thanks again for your contribution!

jaxopt/_src/lbfgs.py
froystig: If we want them as explicit scalar arrays, rather than Python scalars, then `np.array(False)` where `np` is original numy. If we moreover want them on device for some reason, then `jnp.asarray(False)`.

mblondel: Here the main thing we want to be careful of is that `init_state` and `update` output states with the same types. Otherwise, it triggers recompilations.

jaxopt/_src/lbfgs.py
mblondel: I'm confused. I thought `allow_failed_linesearch == True` was JAXopt's behavior? (since the optimization loop keeps going even in the case of failed line search).

zaccharieramzi: Ah it's because there is a confusion about the naming. Originally I had called it `allow_linesearch_failure` which meant that we authorize the linesearch failure to be passed on. Now with the new name I understand why there is some confusion. Maybe we can just change the name?

mblondel: My opinion is that I would keep `stop_if_linesearch_fails=False` by default for the time being because we already have users relying on LBFGS as (inner) solver and we can't take the risk to break their code/tests without any warning. If there is strong evidence that stopping is the right thing to do and this should be the default, we need to use a deprecation cycle to warn that the default value is going to change in release 0.x

jaxopt/_src/lbfgs.py
froystig: I believe we're fine with just a plain literal `False`:
```python
>>> import jax
>>> import jax.numpy as jnp
>>> @jax.jit
... def f(x):
...   print('compiling')
...   return x
... 
>>> f(False)
compiling
DeviceArray(False, dtype=bool)
>>> f(jnp.asarray(False))
DeviceArray(False, dtype=bool)
```

jaxopt/_src/lbfgs.py
zaccharieramzi: Ok for now it is `stop_if_linesearch_fails=False`, keeping it as is.

jaxopt/_src/lbfgs.py
zaccharieramzi: this would mean rewriting the unit tests that check the return types, so maybe it's for a different PR?

mblondel: Yes, I agree. Let's squash the commits and merge this PR.

zaccharieramzi:
@mblondel squashed the commits, and tests are green locally, you can merge when green in CI.

Issue #329, opened 2022-10-19
Issue with has_aux in ArmijoSGD


adamian98:
In the update function for ArmijoSGD, the value of aux is currently not updated:
```
next_state = ArmijoState(iter_num=state.iter_num+1,
                             error=error,
                             value=f_next,
                             aux=state.aux,
                             stepsize=stepsize,
                             velocity=next_velocity)
```

I think that in order to update the value of aux, the fourth line should be `aux=aux` instead of `aux=state.aux`.

mblondel:
Good catch, agreed!

Algue-Rythme:
Fix by #343

PR #343, opened 2022-11-12
Fix missing `aux` in ArmijoSGD #329

Algue-Rythme:
Missing `aux` in #329


mblondel (approved):
LGTM, thank you @Algue-Rythme

PR #328, opened 2022-10-15
Fix missing args,kwargs in resnet example

Algue-Rythme:
Fix bug introduced in https://github.com/google/jaxopt/commit/02bfdc2e2e9947611083b83b7bd1fc6783e3570d

PR #352, opened 2022-12-09
FIX: broken example resnet_flax

fabianp:
The call to init_state was raising an exception. Not sure if the API has recently change or whether it was always been broken ...


Algue-Rythme:
It has always been broken I think, even on the `.py` file: see #328

fabianp:
Thanks @Algue-Rythme ! do you think this fix is correct?

Algue-Rythme:
Yes I guess so!

PR #353, opened 2022-12-10
Remove FLAX resnet example (duplicate in notebook example)

fabianp:
This example was duplicate between the examples and notebooks directory. I think its best to have it just on one directory, to avoid duplicate errors, as we realized in https://github.com/google/jaxopt/pull/352


Algue-Rythme (approved):
LGTM!

PR #354, opened 2022-12-12
Added a new API allowing to warm start the inverse Hessian approximation in LBFGS

### LINKED ISSUE #351: Enable warm-starting the hessian approximation in L-BFGS
Currently one can only provide an initial estimate of the solution, enable warm start of the iterates.
But for quasi-Newton methods, it can also be a good idea to provide initial estimates of the hessian approximation, typically when solving multiple time a similar problem.

This was for example done in HOAG by @fabianp (see https://github.com/fabianp/hoag/blob/master/hoag/hoag.py#L109).

I am willing to implement this in the next few weeks.

As I know it is of interest to them as well, cc-ing @marius311 and @mblondel

zaccharieramzi:
This fixes #351 .

@mblondel I couldn't use your suggestion of creating a new type of init `LBFGSInit` because the `init_params` variable is used for both `init_state` and `update`. Therefore I would have had to add case distinctions in the 2 functions which seemed unreasonable.
Rather I took the approach I saw in some other iterative solvers which was to add an extra keyword argument to `init_state`, `update` and `_value_and_grad_fun`.

I added a test to make sure that this runs, but I am not sure whether we need to add a test to make sure that it improves some cases. I also don't know whether we should test that differentiation is ok.

### DIFF
--- a/jaxopt/_src/base.py
+++ b/jaxopt/_src/base.py
@@ -185,7 +185,10 @@ class IterativeSolver(Solver):
   # search as an iterative solver, but for this reason and others
   # (automatic implicit diff) we should consider having it not be one.
   def _make_zero_step(self, init_params, state) -> OptStep:
-     return OptStep(params=init_params, state=state)
+    if isinstance(init_params, OptStep):
+      return OptStep(params=init_params.params, state=state)
+    else:
+      return OptStep(params=init_params, state=state)
 
   def _run(self,
            init_params: Any,
@@ -378,3 +381,4 @@ class IterativeLineSearch(IterativeSolver):
 
     return super()._run(init_stepsize, params, value, grad, descent_direction,
                         *args, **kwargs)
+
--- a/jaxopt/_src/lbfgs.py
+++ b/jaxopt/_src/lbfgs.py
@@ -257,17 +257,33 @@ class LBFGS(base.IterativeSolver):
     Returns:
       state
     """
+    if isinstance(init_params, base.OptStep):
+      # `init_params` can either be a pytree or an OptStep object
+      state_kwargs = dict(
+        s_history=init_params.state.s_history,
+        y_history=init_params.state.y_history,
+        rho_history=init_params.state.rho_history,
+        gamma=init_params.state.gamma,
+        iter_num=init_params.state.iter_num,
+        stepsize=init_params.state.stepsize,
+      )
+      init_params = init_params.params
+      dtype = tree_single_dtype(init_params)
+    else:
+      dtype = tree_single_dtype(init_params)
+      state_kwargs = dict(
+        s_history=init_history(init_params, self.history_size),
+        y_history=init_history(init_params, self.history_size),
+        rho_history=jnp.zeros(self.history_size, dtype=dtype),
+        gamma=jnp.asarray(1.0, dtype=dtype),
+        iter_num=jnp.asarray(0),
+        stepsize=jnp.asarray(self.max_stepsize, dtype=dtype),
+      )
     (value, aux), grad = self._value_and_grad_with_aux(init_params, *args, **kwargs)
-    dtype = tree_single_dtype(init_params)
-    return LbfgsState(iter_num=jnp.asarray(0),
-                      value=value,
+    return LbfgsState(value=value,
                       grad=grad,
-                      stepsize=jnp.asarray(self.max_stepsize, dtype=dtype),
                       error=jnp.asarray(jnp.inf),
-                      s_history=init_history(init_params, self.history_size),
-                      y_history=init_history(init_params, self.history_size),
-                      rho_history=jnp.zeros(self.history_size, dtype=dtype),
-                      gamma=jnp.asarray(1.0, dtype=dtype),
+                      **state_kwargs,
                       aux=aux,
                       failed_linesearch=jnp.asarray(False))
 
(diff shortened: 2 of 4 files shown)


mblondel:
> I couldn't use your suggestion of creating a new type of init LBFGSInit because the init_params variable is used for both init_state and update. Therefore I would have had to add case distinctions in the 2 functions which seemed unreasonable.

I'm not sure I see why that would be an issue. What I propose is that from now on, `init_params` can be either an array/pytree or a named tuple. Actually, let's just call the new named tuple `jaxopt.Initialization`, since the Hessian could be useful potentially in other solvers. What's wrong with using `if isintance(init_params, jaxopt.Initialization):` within `init_state` and `update`?

The extra arguments of `run` are to pass extra arguments to the objective function. Those are parameters w.r.t. which we want to differentiate the solution. We should avoid adding new arguments. I don't think we want to differentiate w.r.t. the Hessian initialization.

mblondel:
> In some cases the extra arguments of run are not necessarily passed only to the objective function

I agree that the handling of prox hyper-parameters is not ideal but the rule is that `run` and `optimality_fun` should have the same signature.

jaxopt/_src/base.py
```diff
@@ -378,3 +378,8 @@ def run(self,
 
     return super()._run(init_stepsize, params, value, grad, descent_direction,
                         *args, **kwargs)
+
+
+class Initialization(NamedTuple):
+  params: Any
+  state: Any
```
mblondel, line 385: This is exactly the same as `base.OptStep` so I think we should use it instead of introducing `Initialization`. So basically your change allows to resume optimization from any given (point, state) tuple with `run`, i.e., without a manual for loop. CC @fabianp @fllinares

fabianp: It seems indeed a good idea to reuse OptStep here

zaccharieramzi: Agreed, do you think then it can be made into a more general change rather than focused on LBFGS?
Anyway updating now with OptStep

jaxopt/_src/lbfgs.py
```diff
@@ -243,6 +243,12 @@ def _cond_fun(self, inputs):
     # either failed linesearch is disallowed or linesearch hasn't failed.
     return (state.error > self.tol) & jnp.logical_or(not self.stop_if_linesearch_fails, ~state.failed_linesearch)
 
+  def _make_zero_step(self, init_params, state) -> base.OptStep:
```
mblondel, line 246: Let's move it to base.py

jaxopt/_src/lbfgs.py
```diff
@@ -256,19 +262,34 @@ def init_state(self,
...
     if self.has_aux:
       _, aux = self.fun(init_params, *args, **kwargs)
     else:
       aux = None
-    dtype = tree_single_dtype(init_params)
-    return LbfgsState(iter_num=jnp.asarray(0),
-                      value=jnp.asarray(jnp.inf),
+    return LbfgsState(value=jnp.asarray(jnp.inf),
                       stepsize=jnp.asarray(self.max_stepsize, dtype=dtype),
                       error=jnp.asarray(jnp.inf),
```
mblondel, line 291: Shouldn't we initialize `value`, `stepsize` and `errror` too?

jaxopt/_src/lbfgs.py
```diff
@@ -256,19 +262,34 @@ def init_state(self,
     Returns:
       state
     """
+    if isinstance(init_params, base.OptStep):
+      # this allows to maintain backward compatibility with the previous API
+      hessian_approx_kwargs = dict(
```
mblondel, line 267: `iter_num` is not Hessian-related so maybe just rename the variable to `kwargs`

mblondel (commented):
Thanks a lot @zaccharieramzi!

I propose we merge this PR first and we can implement the same API in other solvers separately. We can also document it in the narrative doc.

jaxopt/_src/lbfgs.py
zaccharieramzi: sure!

jaxopt/_src/lbfgs.py
zaccharieramzi: I think it's a good question. 
For `value` and `error` there is not really any interest in doing so, it's not going to affect how the algorithm runs.
For stepsize it's a bit different since the linesearch is based off of that.
I would say it's probably best to give the user the choice so initialize `stepsize` with the value contained in the state, which the user can always change, and make sure it's well documented.

jaxopt/_src/lbfgs.py
zaccharieramzi: since I will initialize `stepsize` I will change it to `kwargs` not matter what.

zaccharieramzi:
@mblondel when you mention the narrative doc, do you mean [this page](https://jaxopt.github.io/stable/root_finding.html)?
Happy to indeed add some content about L-BFGS, maybe in a subsequent PR to keep things tight, wdyt?

jaxopt/_src/lbfgs.py
```diff
@@ -256,19 +256,35 @@ def init_state(self,
     Returns:
       state
     """
+    if isinstance(init_params, base.OptStep):
+      # this allows to maintain backward compatibility with the previous API
```
mblondel, line 260:
```suggestion
      # `init_params` can either be a pytree or an OptStep object
```
I fear that your comment regarding "the old" API will be difficult to understand in a few years.

Also, can you update the docstring of `run` to mention that OptStep objects are supported for `init_params`?

mblondel (approved):
LGTM, apart from my minor 2 comments below. Let's merge after squashing commits.

zaccharieramzi:
all commits squashed

zaccharieramzi:
@mblondel merged main, re-tested and all green, and squashed commits

PR #357, opened 2022-12-14
Add pylintrc to top level directory

### LINKED ISSUE #355: Include JAX's pylintrc file
I find it useful to use a python linter on the project, but these usually don't come by default with Google's python style. JAX for example includes a pylintrc file in the top level so that all devs  have the same config: https://github.com/google/jax/blob/main/pylintrc

I propose we do the same and include this file in JAXopt. Thoughts?

fabianp:
fixes #355


mblondel (approved):
LGTM, thank you

PR #346, opened 2022-11-17
Custom loop pmap example

fllinares:
A MWE of how `jax.pamp` can be used in JAXopt (see issue #333 )

PR #358, opened 2022-12-20
Custom loop pjit example

fllinares:
A MWE of how `jax.experimental.pjit` can be used in JAXopt (see also PR #346).

NOTE: `jax.experimental.pjit` is not yet supported in Colab. However, this example illustrates how users with access to Google Cloud TPUs may use `jax.experimental.pjit` in combination with JAXopt solvers.


docs/notebooks/distributed/custom_loop_pjit_example.md
```diff
@@ -0,0 +1,375 @@
...
+  elapsed: 41809
+  status: ok
+  timestamp: 1671545341521
+  user:
+    displayName: Felipe Llinares
+    userId: 01655756739108476525
+  user_tz: -60
+id: JJuP-Wz_MBeJ
+---
+import functools
```
fabianp, line 95: unused

docs/notebooks/distributed/custom_loop_pjit_example.md
```diff
@@ -0,0 +1,375 @@
...
+import time
+from typing import Any, Callable, Tuple, Union
+
+from absl import app
```
fabianp, line 99: both app and flags seem to be unused

docs/notebooks/distributed/custom_loop_pjit_example.md
```diff
@@ -0,0 +1,375 @@
...
+from absl import flags
+
+from flax import jax_utils
+from flax.training import common_utils
```
fabianp, line 103: both jax_utils and common_utils are unused

docs/notebooks/distributed/custom_loop_pjit_example.md
```diff
@@ -0,0 +1,375 @@
...
+NUM_CLASSES = 10  #@param{type:"integer"}
+#@markdown The stepsize for the optimizer (set to 0.0 to use line search):
+STEPSIZE = 0.0  #@param{type:"number"}
+#@markdown The line search approach (either `'zoom'` or `backtracking`), ignored if `STEPSIZE > 0.0`:
+LINESEARCH = 'zoom'  #@param{type:"string"}
+```
+
++++ {"id": "bZmsaMOcONjQ"}
+
+# Imports and TPU setup
```
fabianp, line 81: I usually add 
"
%%capture
%pip install jaxopt
"
as the first cell to install jaxopt in the examples (since its not by default in colab). Although here its less important as its not meant to be run (today) on colab, I would still add it so it installs jaxopt if its not available in the system and for consistency

fabianp (changes requested):
Thanks Felipe!

docs/notebooks/distributed/custom_loop_pjit_example.md
fllinares: Thanks, done!

docs/notebooks/distributed/custom_loop_pjit_example.md
fllinares: Thanks, done!

docs/notebooks/distributed/custom_loop_pjit_example.md
fllinares: Idem.

docs/notebooks/distributed/custom_loop_pjit_example.md
fllinares: Good point. I was doubting what to do, since for now only running on GCP is supported, and that magic will not be useful in that case. But it's better to future-proof the notebook.

fabianp (approved):
thanks! LGTM

fabianp:
BTW I really like how you organized the notebook parameters

For some context, it looks like this on a notebook:
image

@mblondel : what do you think of adopting this convention on all notebooks? it feels less clunky than the NamedTuple option

fabianp:
merged through #365

Issue #202, opened 2022-04-13
make notebook (colab) examples work on TPUs


fabianp:
these are the examples on docs/notebooks/

Probably it's necessary to add this line when running on TPUs https://github.com/google/jax#pip-installation-colab-tpu

I think this would be a good first issue

Samad999777:
Hey @fabianp is this still open for contribution ?

fabianp:
yes!

Samad999777:
Great, I would like to work on it

fabianp:
all yours :-)

Samad999777:
I actually have to reproduce the already given examples in colab notebooks using TPU?

fabianp:
yes, the goal is that the examples run and produce similar results using TPU instead of GPU. Some might work out of the box, some might fail (nobody knows because nobody has tried yet)

Samad999777:
Alright, I will keep you posted about it. You can assign this issue to me.

Samad999777:
Hey @fabianp I have a colab notebook working for one example. Should i create a pull request so you can check it out?

fabianp:
Yes please :-)

fabianp:
@Samad999777 : any updates on this :-) ?

fabianp:
There's no rush, I was reaching out mostly in case you needed some help.
Again, there's no urgency in this

Samad999777:
@fabianp thanks alot, hopefully we will wrap it up this week.

rosspeili:
Anyone solved this?

PR #361, opened 2023-01-04
Add Colab TPU initialization

ayush-1506:
Using TPUs with JAX on colab requires an extra step. I ran it for a couple of examples with TPU acceleration on colab and it seems to work.


ayush-1506:
Coincidentally, also fixes #202

ayush-1506:
Also fixes #202 .

ayush-1506:
Without TPU, it's going to throw a KeyError (`COLAB_ADDR_TPU not found`).

ayush-1506:
@fabianp How about just catching this error and logging something like `TPU not detected, running without it` if TPU isn't detected?

fabianp:
yes! that's a good idea. Could you please modify the code to do this? Thanks!

ayush-1506:
@fabianp should work now.

fabianp:
excellent! We’re almost there. A couple of things to do before merging:

0. pull from the main repository. I think there’s a new notebook in the folder notebooks/perturbed_optimizers . Please add the same trick to allow it to run on TPUs.
1. sync the markdown jupyter notebooks as mentioned here: https://jaxopt.github.io/stable/developer.html#syncing-notebooks
2. squash all the commits in this pull request so this pull request contains a single commit.

Thanks again for the contribution!

fabianp:
thanks @ayush-1506 . One last thing. Please add a a comment before the `import jax.tools.colab_tpu` line to explain what is happening? something like `# activate TPUs if available`. After that we're good to merge

fabianp (approved):

fabianp:
all done and merged. Thanks @ayush-1506 !

PR #367, opened 2023-01-10
Added possibility for line search methods to return a potential auxiliary value in their results

zaccharieramzi:
Linesearch methods affected : Backtracking linesearch and Zoom linesearch.

This was was also used in order to correct the way auxiliary values are handled in the BFGS, L-BFGS and Nonlinear CG solvers where these methods are used.

I still need to add tests on the matching between the auxiliary value and the parameters returned.

### DIFF
--- a/jaxopt/_src/backtracking_linesearch.py
+++ b/jaxopt/_src/backtracking_linesearch.py
@@ -40,6 +40,7 @@ class BacktrackingLineSearchState(NamedTuple):
   grad: Any
   num_fun_eval: int
   num_grad_eval: int
+  aux: Optional[Any] = None
 
 
 @dataclass(eq=False)
@@ -53,6 +54,9 @@ class BacktrackingLineSearch(base.IterativeLineSearch):
     value_and_grad: if ``False``, ``fun`` should return the function value only.
       If ``True``, ``fun`` should return both the function value and the
       gradient.
+    has_aux: if ``False``, ``fun`` should return the function value only.
+      If ``True``, ``fun`` should return a pair ``(value, aux)`` where ``aux``
+      is a pytree of auxiliary values.
 
     condition: either "armijo", "goldstein", "strong-wolfe" or "wolfe".
     c1: constant used by the (strong) Wolfe condition.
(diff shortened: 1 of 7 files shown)


jaxopt/_src/nonlinear_cg.py
```diff
@@ -180,23 +182,29 @@ def update(self,
...
+      t = new_stepsize.astype(tree_single_dtype(params))
+      new_params = tree_add_scalar_mul(params, t, descent_direction)
 
     else:
       raise ValueError("Invalid name in 'linesearch' option.")
 
-    new_params = tree_add_scalar_mul(params, new_stepsize, descent_direction)
-    (new_value, new_aux), new_grad = self._value_and_grad_with_aux(new_params,
-                                                                   *args,
-                                                                   **kwargs)
```
zaccharieramzi, line 198: This extra call was only needed to compute values that were already accessible in the outputs of the line search methods, except for `new_aux`.
Now that `new_aux` is available in the output of the linesearch solvers, we can get rid of this call.

Note that this differs from the BFGS and L-BFGS solvers because these quasi-Newton solvers support a stepsize API.

zaccharieramzi:
@mblondel I added the tests to check for the auxiliary value, all green in local. This can be reviewed.

jaxopt/_src/backtracking_linesearch.py
```diff
@@ -53,6 +54,7 @@ class BacktrackingLineSearch(base.IterativeLineSearch):
     value_and_grad: if ``False``, ``fun`` should return the function value only.
       If ``True``, ``fun`` should return both the function value and the
       gradient.
+    has_aux: if ``False``, ``fun`` should return the function value only.
```
mblondel, line 58: Mention the True case too.

mblondel (commented):
Good work! LGTM apart from my minor comment below.

jaxopt/_src/backtracking_linesearch.py
zaccharieramzi: sure done

zaccharieramzi:
@mblondel squashed all commits as well

PR #370, opened 2023-01-10
Got rid of duplicate LBFGS/BFGS function calls in the no-linesearch case

zaccharieramzi:
In the no-linesearch case, as was noted by [this comment](https://github.com/google/jaxopt/blob/ac2d1e6f3797aed18498c19b1294307056472170/jaxopt/_src/lbfgs.py#L351), L-BFGS was doing an extra computation of function value and gradient.
Same for BFGS.

This PR also adds a test to check the number of calls.


tests/common_test.py
```diff
@@ -60,7 +60,7 @@ def fixed_point_fun(params, *args, **kwargs):
 
       hash(solver)  # Checks that hash works.
       update = jax.jit(solver.update)
-      state0 = solver.init_state(params0)
+      state0 = solver.init_state(params0, data=data)
```
zaccharieramzi, line 63: this had not been spotted before because the init state methods were not computing the value function most of the solvers not requiring it in the absence of auxiliary value

zaccharieramzi:
Currently tests fail because of projection_test. Indeed under the hood, this uses LBFGS which is affected by our changes.
From my debugging it seems that the main reason for this change of behaviour is that we are not recomputing the gradient after the call to zoom linesearch.

Indeed, in some cases we have the gradient taken directly from the linesearch is not the one that you can compute from the new parameters. This happens when the linesearch fails, and that the failure is not handled. 

I will solve this in a separate PR.

zaccharieramzi:
Since the fix on the linesearch failed state was also applicable to nonlinear CG I implemented it in this case as well.

@mblondel all tests passing in local (although for the nonlinear CG one I did observe the arbitrary_pytree test go red at some point, certainly because the randomness is not well fixed, but green for all other runs I did).

This is ready for review once https://github.com/google/jaxopt/pull/367 is merged.

tests/bfgs_test.py
```diff
@@ -113,5 +115,24 @@ def test_multiclass_logreg(self, linesearch):
...
+      global n_calls
+      n_calls += 1
+      return sum(100.0*(x[1:] - x[:-1]**2.0)**2.0 + (1 - x[:-1])**2.0)
+
+    with jax.disable_jit(True):
+      x0 = jnp.zeros(2)
+      bfgs = BFGS(fun=fun, tol=1e-7, maxiter=n_iter, implicit_diff=False, stepsize=1e-3)
+      x1, _ = bfgs.run(x0)
+
+    assert n_calls == n_iter + 1
```
mblondel, line 134: Use self.assertEqual (the error message will be more intelligible)

tests/lbfgs_test.py
```diff
@@ -277,6 +279,25 @@ def f(x):
...
+      global n_calls
+      n_calls += 1
+      return sum(100.0*(x[1:] - x[:-1]**2.0)**2.0 + (1 - x[:-1])**2.0)
+
+    with jax.disable_jit(True):
+      x0 = jnp.zeros(2)
+      lbfgs = LBFGS(fun=fun, tol=1e-7, maxiter=n_iter, implicit_diff=False, stepsize=1e-3)
+      x, _ = lbfgs.run(x0)
+
+    assert n_calls == n_iter + 1
```
mblondel, line 298: Same

mblondel (approved):
LGTM apart from my minor comments below. Thank you @zaccharieramzi

tests/bfgs_test.py
zaccharieramzi: done

tests/lbfgs_test.py
zaccharieramzi: done

mblondel (approved):
LGTM, thanks @zaccharieramzi

PR #375, opened 2023-01-11
Dtype consistency between input and output of update method.

mblondel:


fllinares (approved):
Thanks a lot Mathieu, looks great!!

PR #377, opened 2023-01-11
API reference for isotonic regression

fabianp:


jaxopt/isotonic.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2021 Google LLC
```
mblondel, line 1:
```suggestion
# Copyright 2023 Google LLC
```

mblondel (approved):
LGTM, apart from year in header file. Thank you!

Issue #381, opened 2023-01-19
Retrieving inverse hessian when using `ScipyMinimize('BFGS')`


aymgal:
Hi,
Is there any way to retrieve the `hess_inv` (approximated inverse Hessian) after running the BFGS optimizer? In the original scipy's implementation, the matrix is stored in the return `OptimizeResult` object.

mblondel:
I don't think so, but a patch would be welcome.

aymgal:
Ok, I proposed a patch with #382

PR #382, opened 2023-01-19
Add inverse hessian approximation to the returned state

aymgal:
The `state` returned after running `ScipyMinimize` now contains the approximated inverse Hessian matrix (if the optimizer supports it, e.g. `'BFGS'`). The inverse Hessian matrix gives an approximation of the covariance matrix.
A small test has been added to check against original scipy. 
The goal is to solve issue #381.

### DIFF
--- a/jaxopt/_src/scipy_wrappers.py
+++ b/jaxopt/_src/scipy_wrappers.py
@@ -36,6 +36,7 @@ from typing import Union
 import jax
 import jax.numpy as jnp
 import jax.tree_util as tree_util
+from jax.config import config
 
 from jaxopt._src import base
 from jaxopt._src import implicit_diff as idf
@@ -66,7 +67,7 @@ class LbfgsInvHessProductPyTree(LbfgsInvHessProduct):
     if sk.shape != yk.shape or sk.ndim != 2:
       raise ValueError('sk and yk must have matching shape, (n_corrs, n)')
     n_corrs, n = sk.shape
-    self.dtype = jnp.float64
+    self.dtype = jnp.float64 if config.jax_enable_x64 is True else jnp.float32
     self.shape = (n, n)
     self.sk = sk
     self.yk = yk
--- a/jaxopt/_src/scipy_wrappers.py
+++ b/jaxopt/_src/scipy_wrappers.py
@@ -41,9 +41,50 @@ from jaxopt._src import base
 from jaxopt._src import implicit_diff as idf
 from jaxopt._src import projection
 from jaxopt._src.tree_util import tree_sub
+from jax.tree_util import register_pytree_node_class
 
 import numpy as onp
 import scipy as osp
+from scipy.optimize import LbfgsInvHessProduct
+
+
+@register_pytree_node_class
+class LbfgsInvHessProductPyTree(LbfgsInvHessProduct):
+  """
+  Registers the LbfgsInvHessProduct object as a PyTree.
+  This object is typically returned by the L-BFSG-B optimizer to efficiently 
+  store the inverse of the Hessian matrix evaluated at the best-fit parameters. 
+  """
+
+  def __init__(self, sk, yk):
+    """
+    Construct the operator.
+    This is the same constructor as the original LbfgsInvHessProduct class,
+    except that numpy has been replaced by jax.numpy and no call to the 
+    numpy.ndarray constuctor is performed.
+    """
+    if sk.shape != yk.shape or sk.ndim != 2:
+      raise ValueError('sk and yk must have matching shape, (n_corrs, n)')
+    n_corrs, n = sk.shape
+    self.dtype = jnp.float64
+    self.shape = (n, n)
+    self.sk = sk
+    self.yk = yk
+    self.n_corrs = n_corrs
+    self.rho = 1 / jnp.einsum('ij,ij->i', sk, yk)
+
+
+  def __repr__(self):
+      return "LbfgsInvHessProduct(sk={}, yk={})".format(self.sk, self.yk)
+
+  def tree_flatten(self):
+      children = (self.sk, self.yk)
+      aux_data = None
+      return (children, aux_data)
+
+  @classmethod
+  def tree_unflatten(cls, aux_data, children):
+      return cls(*children)
 
 
 class ScipyMinimizeInfo(NamedTuple):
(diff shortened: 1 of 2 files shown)


jaxopt/_src/scipy_wrappers.py
```diff
@@ -312,10 +313,19 @@ def scipy_fun(x_onp: onp.ndarray) -> Tuple[onp.ndarray, onp.ndarray]:
                                 options=self.options)
 
     params = tree_util.tree_map(jnp.asarray, onp_to_jnp(res.x))
+    if hasattr(res, 'hess_inv'):
+        if isinstance(res.hess_inv, onp.ndarray):
+            hess_inv = res.hess_inv
+        else:
+            hess_inv = res.hess_inv.todense()
```
mblondel, line 320: Silently densifying sparse matrices is not ideal. This could blow up memory... It may be better to raise an exception for the time being.

mblondel (commented):
LGTM apart from the remark below

jaxopt/_src/scipy_wrappers.py
mblondel: JAX has an experimental sparse module https://jax.readthedocs.io/en/latest/jax.experimental.sparse.html

aymgal: Indeed, I understand why it's a problem to always densify it.
However, most of the time (at least in the test routines), the returned object is not a numpy array, but a `LbfgsInvHessProduct` instance (see [here](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.LbfgsInvHessProduct.html))... and I could not find a way to extract the matrix as sparse matrix.
Maybe just return the `LbfgsInvHessProduct` as it is? There is no JAX version of it, if I'm correct. I tried to get the hessian out of the operator without turning it into a dense matrix first (e.g. via a product with JAX's sparse matrix, but it returns a numpy dense array). 
Do you have a specific idea of how to use the sparse module from JAX in this context?

mblondel: Sorry for the late reply. From what you said, simply adding the `LbfgsInvHessProduct` object (without converting it to an array) to the state seems like the way to go. What do you think?

aymgal: @mblondel The problem is that this object cannot be cast by default to a JAX object. If I directly return the object itself, I get this error after running the optimizer `'L-BFGS-B'`:
```
TypeError: Value <2x2 LbfgsInvHessProduct with dtype=float64> with type <class 'scipy.optimize.lbfgsb.LbfgsInvHessProduct'> is not a valid JAX type
```
There should be a way to extract the matrix while keeping it as sparse in memory, but I can't figure out how. Or to wrap the scipy's `LbfgsInvHessProduct` object to a JAX-friendly one, but I am not sure how to do this either.

For `'BFGS'` there is no such problem, since the returned hessian is an actual matrix.

aymgal: Or what about allowing the user to decide if the hess_inv should be returned or not? By default it would not be, but if the user knows that memory won't be a problem or if the number of parameters is low, it densifies the matrix and returns it.

mblondel: I feel it should be possible to wrap `LbfgsInvHessProduct` into a pytree (see `RegisteredSpecial2` example [here](https://jax.readthedocs.io/en/latest/pytrees.html))

@froystig Any recommendation on this?

froystig: Resources on custom pytrees registration:
* https://jax.readthedocs.io/en/latest/pytrees.html
* https://jax.readthedocs.io/en/latest/jax-101/05.1-pytrees.html

aymgal:
Thanks @mblondel and @froystig for the tips. I added a new class to register the scipy's LbfgsInvHessProduct class as a pytree. It seems to work now, I also updated the test script. Let me know what you think!

tests/scipy_wrappers_test.py
```diff
@@ -601,6 +601,46 @@ def wrapper(box_len):
...
+    true_sol = [0.5, -1.2]
+    # define a very simple objective
+    self.objective = lambda x: (x[0] - true_sol[0])**2. + (x[1] - true_sol[1])**2.
+    # analytical derivative of the objective function
+    self.derivative = lambda x: [(x[0] - true_sol[0]) * 2., (x[1] - true_sol[1]) * 2.]
+    # define the starting point as a random sample from the domain
+    r_min, r_max = -5.0, 5.0
+    onp.random.seed(6574)
+    self.x0 = r_min + onp.random.rand(2) * (r_max - r_min)
+    #self.x0 = [1., 1.]
```
mblondel, line 618: Remove

tests/scipy_wrappers_test.py
```diff
@@ -601,6 +601,46 @@ def wrapper(box_len):
...
+                                      method='BFGS', jac=self.derivative)
+    # print(res_scipy.x, res_scipy.hess_inv)
+    # raise
+    # run the same algorithm via jaxopt
+    res_jaxopt = ScipyMinimize(method='BFGS', fun=self.objective).run(self.x0)
+    # compare the two returned inverse Hessian matrices
+    self.assertAllClose(res_scipy.hess_inv, res_jaxopt.state.hess_inv)
+    # onp.testing.assert_almost_equal(res_scipy.hess_inv, 
+    #                                 res_jaxopt.state.hess_inv, 
+    #                                 decimal=6)
```
mblondel, line 632: Remove commented assertion

mblondel (commented):
Apologies for the late reply. LGTM apart from some left-over comments.

By the way, could you squash the commits? Thanks!

aymgal:
Thanks for the feedback, I removed the commented lines and squashed the 5 commits.

mblondel:
There is a failure on line 51 `AttributeError: module 'scipy' has no attribute 'optimize'`. Not sure what's going on, this line should work.

aymgal:
@mblondel That was strange indeed, it was working on my machine. I changed the way in import `LbfgsInvHessProduct`, let's see if it works.
I also noticed that other tests that I missed before were failing, in relation to the LBFGS optimizer. Now it should work, hopefully.

aymgal:
@mblondel a test is still failing, but it does not seem to be related in any way to this PR. The failed test is `bisection_test.py`, which never relates to the `scipy_wrappers` module on which I've worked on.

froystig:
I believe that failure (indicating a jax tracer leak) is one that is currently known, also on the `main` branch.

froystig:
> I believe that failure (indicating a jax tracer leak) is one that is currently known, also on the `main` branch.

#412 ought to fix that failure

mblondel:
If you merge main in your branch, this should now work. Thank you for your patience.

mblondel (approved):

PR #379, opened 2023-01-15
enable `pip` caching in CI + add `pyproject.toml`

SauravMaheshkar:
Changes made by this PR can be summarized as follows:

- Use an updated version of `actions/setup-python` so as to allow the caching of `pip`.

PR #411, opened 2023-03-15
enable `pip` caching in CI + add `pyproject.toml`

SauravMaheshkar:
Port over changes from #379.

Changes made by this PR can be summarized as follows:

- Use an updated version of `actions/setup-python` so as to allow the caching of `pip`.
- switch to `pyproject.toml` as a build backend.


SauravMaheshkar:
@mblondel Request for Review

SauravMaheshkar:
It seems like the build is passing, but the tests are failing. I don't think the tests are related to the changes proposed by this PR. Infact the step changed by this PR ("Build") works.

mblondel:
The CI failures should be fixed when #412 is merged.

I'm am not familiar with poetry. What would be the benefits? I see that you removed our setup.py file. This will break our [workflow](https://github.com/google/jaxopt/blob/main/.github/workflows/publish_to_pypi.yml).

I'm +1 for the caching but less convinced for poertry. It would be great if you could submit the caching as a separate PR.

SauravMaheshkar:
I'm not proposing we switch to using poetry. We'll still use pip as the package manager, it's just that the metadata will be specified in the `pyproject.toml` file. And I've updated the workflow as well so that shouldn't break.

PR #413, opened 2023-03-16
Added Option to Reduce Variance of Jacobian of Perturbed Argmax

LawrenceMMStewart:
The jacobian of the perturbed argmax [Proposition 3.1 Berthet et al.](https://arxiv.org/pdf/2002.08676.pdf)  explodes as sigma tends to 0. If the noise distribution is has zero mean (such as `perturbations.Normal`), then it is possible to avoid this problem by replacing y(theta + sigma Z) with y(theta + sigma Z)  - y(theta) in the monte-carlo estimator, as described in [Le Lidec 2021](https://proceedings.neurips.cc/paper/2021/file/ab233b682ec355648e7891e66c54191b-Paper.pdf) and [Bach 2023](https://www.di.ens.fr/~fbach/ltfp_book.pdf).

This PR implements this minor change by adding a boolean variable `has_zero_mean` to the noise distributions in `perturbations` as well as adding another optional boolean argument to `perturbations.make_perturbed_argmax`, called `reduce_variance`. If both of these booleans are true, then the Jacobian is instead calculated in the above manner, with reduced variance.

I have provided [an example notebook](https://github.com/LawrenceMMStewart/make_perturbed_argmax-variance-reduction/blob/main/Jaxopt-Variance-Reduction.ipynb), which demonstrates the functionality of the PR on a simple differentiable ranking examples similar to that of the existing jaxopt example gallery. 

This implementation can be further optimized. Instead of calculating `output = argmax_fun(inputs)` in the function `pert_jvp_reduce_variance`, we can use the fact that this value was already recorded in the forward pass by the function `forward_pert`. However, I am unsure how one goes about accessing the output of a forward pass in a `custom_jvp` function.

### DIFF
--- a/jaxopt/_src/perturbations.py
+++ b/jaxopt/_src/perturbations.py
@@ -21,9 +21,6 @@ import jax.numpy as jnp
 class Normal:
   """Normal distribution."""
 
-  def __init__(self):
-      self.has_zero_mean = True
-
   def sample(self,
              seed: jax.random.PRNGKey,
              sample_shape: Tuple[int]) -> jax.Array:
@@ -35,9 +32,6 @@ class Normal:
 
 class Gumbel:
   """Gumbel distribution."""
-
-  def __init__(self):
-      self.has_zero_mean = False
 
   def sample(self,
              seed: jax.random.PRNGKey,
@@ -64,8 +58,7 @@ def make_perturbed_argmax(argmax_fun: Callable[[jax.Array], jax.Array],
     noise: a distribution object that must implement a sample function and a
       log-pdf of the desired distribution, similar to the examples above.
       Default is Gumbel distribution.
-    reduce_variance : If noise distribution has zero mean use an alternative
-      estimator for the Jacobian which is more stable for small values of sigma.
+    reduce_variance : Option to reduce variance of gradients.
       Default is False.
 
   Returns:
@@ -130,11 +123,11 @@ def make_perturbed_argmax(argmax_fun: Callable[[jax.Array], jax.Array],
         jnp.reshape(tangent, (-1,)))
     return jnp.reshape(tangent_flat, inputs.shape)
 
-  # If noise has zero mean then give option for variance reduction. 
-  if reduce_variance and noise.has_zero_mean:
+  if reduce_variance:
       forward_pert.defjvps(pert_jvp_reduce_variance, None)
   else:
       forward_pert.defjvps(pert_jvp, None)
+
   return forward_pert
 
 
--- a/jaxopt/_src/perturbations.py
+++ b/jaxopt/_src/perturbations.py
@@ -46,7 +46,7 @@ def make_perturbed_argmax(argmax_fun: Callable[[jax.Array], jax.Array],
                           num_samples: int = 1000,
                           sigma: float = 0.1,
                           noise=Gumbel(),
-                          reduce_variance: bool = False):
+                          control_variate: bool = False):
   """Transforms a function into a differentiable version with perturbations.
 
   Args:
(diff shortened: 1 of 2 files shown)


LawrenceMMStewart:
Hello. Thanks @q-berthet you absolutely correct w.r.t to E[score]= 0. I have fetched the most recent changes to jaxopt and updated with the suggestions you made. Let me know intend to add equivalent changes to the function `make_perturbed_fun` and its respective tests, as I am equally happy to do this in the coming week.

@mblondel I have added checks ensure that `reduce_variance=True`  still passes all prior tests relating to `make_perturbed_argmax`. There is also an additional test ensuring that for small sigma, the option `reduce_variance` keeps the jacobian closer to the true value (in the case of F = onehot). 

Let me know if there is anything else to add.

tests/perturbations_test.py
```diff
@@ -107,28 +118,90 @@ def test_argmax(self):
         num_samples=self.num_samples,
         sigma=self.sigma,
         noise=perturbations.Gumbel()))
+
+    # same as pert_argmax_fun but with reduced variance
+    pert_argmax_fun_rv = jax.jit(perturbations.make_perturbed_argmax(
+        argmax_fun=one_hot_argmax,
+        num_samples=self.num_samples,
+        sigma=self.sigma,
+        noise=perturbations.Gumbel()))
```
q-berthet, line 127: same here, missing `reduce_variance=True`

q-berthet (commented):
Thanks Lawrence!

I left some comments, I think something needs to be fixed in the tests.

jaxopt/_src/perturbations.py
```diff
@@ -45,7 +45,8 @@ def log_prob(self, inputs: jax.Array) -> jax.Array:
 def make_perturbed_argmax(argmax_fun: Callable[[jax.Array], jax.Array],
                           num_samples: int = 1000,
                           sigma: float = 0.1,
-                          noise=Gumbel()):
+                          noise=Gumbel(),
+                          reduce_variance: bool = False):
```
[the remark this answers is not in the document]
LawrenceMMStewart: Hi @q-berthet,

Thank you for the corrections (I have just corrected them in the new commit).

You are completely right, `control_variate` is a much better name, instead of `reduce_variance`. Indeed for large `sigma`, there are cases in which no control variate can give better estimations of the true jacobian. I also changed the docstring for the function to explain what is going on a bit clearer. Cheers!
Lawrence

jaxopt/_src/perturbations.py
```diff
@@ -57,6 +58,11 @@ def make_perturbed_argmax(argmax_fun: Callable[[jax.Array], jax.Array],
     noise: a distribution object that must implement a sample function and a
       log-pdf of the desired distribution, similar to the examples above.
       Default is Gumbel distribution.
+    control_variate : Subtract a control variate (the unperturbed argmax) from
```
q-berthet, line 61: I would change this a bit, starting with 'a boolean, indicates whether, ...' to specify the nature - `control_variate` is not a function that substracts.

jaxopt/_src/perturbations.py
```diff
@@ -57,6 +58,11 @@ def make_perturbed_argmax(argmax_fun: Callable[[jax.Array], jax.Array],
...
+      the perturbed argmax in the monte-carlo estimate of the jacobian. For
```
q-berthet, line 62: Monte-Carlo

tests/perturbations_test.py
```diff
@@ -107,28 +119,91 @@ def test_argmax(self):
         num_samples=self.num_samples,
         sigma=self.sigma,
         noise=perturbations.Gumbel()))
+
+    # same as pert_argmax_fun but with reduced variance
```
q-berthet, line 123: same

q-berthet (commented):
Thanks a lot Lawrence, I think that this works. Just highlighted some typos / recommended changes in the comments.

PR #416, opened 2023-03-24
Implemented (Good) Broyden

zaccharieramzi:
This PR adds the implementation of the good Broyden method (denoted in `scipy` as `'broyden1'`).

Some current pending aspects that I would like to discuss (and this is why I am opening this PR):
- I currently have a very hard time getting any test to pass when using float32, as numerical imprecisions cause nan and inf values to appear in my Broyden matrix computations. @mblondel if you have any idea on how to typically resolve those any input is welcome.
- I noticed that the setting of `gamma` is super important in many tests: maybe I should do like scipy and compute it if not set. Btw, unlike LBFGS I didnt find a rule to update it with iterations, so for now it is fixed.
- For now I rely on `jax` autodiff to perform the line search, whereas `scipy` computes an adhoc version of the derivative of `||f(x)||^2` (see [here](https://github.com/scipy/scipy/blob/9562a4a1fd1005eefb7221d1675ef6dcedcf07b8/scipy/optimize/_nonlin.py#L285)). I think it would be nice to compute this adhoc trick only when the function is not differentiable, so I need to find a way to check this nicely.
- similarly to LBFGS, it's not possible to use the unrolled differentiation when linesearch is used. I am wondering whether we should have a more explicit error (atm it's something like `cannot differentiate through while_loop`).
- I have implemented a limited memory version of the broyden method like scipy. However, I currently am not implementing a mechanism to switch to a full matrix if the history size is bigger than the variable dimension (this is also not implemented in LBFGS). Maybe for now a warning to the user could be enough?

Happy to get your thoughts on which of these aspects should be in this PR and which can wait.

### DIFF
--- a/docs/root_finding.rst
+++ b/docs/root_finding.rst
@@ -68,3 +68,49 @@ Scipy wrapper
   :toctree: _autosummary
 
     jaxopt.ScipyRootFinding
+
+
+Broyden's method
+--------------
+
+.. autosummary::
+  :toctree: _autosummary
+
+    jaxopt.Broyden
+
+Broyden's method is an iterative algorithm suitable for nonlinear root equations in any dimension.
+It is a quasi-Newton method (like L-BFGS), meaning that it uses an approximation of the Jacobian matrix
+at each iteration.
+The approximation is updated at each iteration with a rank-one update.
+This makes the approximation easy to invert using the Sherman-Morrison formula, given it does not use too many
+updates.
+One can control the number of updates with the ``history_size`` argument.
+Furthermore, Broyden's method uses a line search to ensure the rank-one updates are stable.
+
+Example::
+
+    import jax.numpy as jnp
+    from jaxopt import Broyden
+
+    def F(x):
+      return x ** 3 - x - 2
+
+    broyden = Broyden(fun=F)
+    print(broyden.run(jnp.array(1.0)).params)
+
+
+For implicit differentiation::
+
+    import jax
+    import jax.numpy as jnp
+    from jaxopt import Broyden
+
+    def F(x, factor):
+      return factor * x ** 3 - x - 2
+
+    def root(factor):
+      broyden = Broyden(fun=F)
+      return broyden.run(jnp.array(1.0), factor=factor).params
+
+    # Derivative of root with respect to factor at 2.0.
+    print(jax.grad(root)(2.0))
--- a/jaxopt/_src/broyden.py
+++ b/jaxopt/_src/broyden.py
@@ -1,0 +1,395 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Limited-memory Broyden method"""
+
+from functools import partial
+
+from typing import Any
+from typing import Callable
+from typing import NamedTuple
+from typing import Optional
+from typing import Union
+
+from dataclasses import dataclass
+
+import jax
+import jax.numpy as jnp
+
+from jaxopt._src import base
+from jaxopt._src.backtracking_linesearch import BacktrackingLineSearch
+from jaxopt.tree_util import tree_map
+from jaxopt.tree_util import tree_vdot
+from jaxopt.tree_util import tree_add_scalar_mul
+from jaxopt.tree_util import tree_scalar_mul
+from jaxopt.tree_util import tree_sub
+from jaxopt.tree_util import tree_l2_norm
+from jaxopt._src.tree_util import tree_single_dtype
+
+
+def matvec(d_history, c_history, x, indices):
+  # CD^T x
+  # here j is the history dimension and i is the space dimension
+  dtx = jnp.einsum('j..., ... -> j', d_history[indices], x)
+  cdtx = jnp.einsum('j..., j -> ...', c_history[indices], dtx)
+  return cdtx
+
+def inv_jacobian_product_leaf(v: jnp.ndarray,
+                              d_history: jnp.ndarray,
+                              c_history: jnp.ndarray,
+                              gamma: float = 1.0,
+                              start: int= 0):
+  # the computation of the jacobian inverse for
+  # the Broyden method
+  # this is done using the notations of DEQ
+  history_size = len(d_history)
+
+  indices = (start + jnp.arange(history_size)) % history_size
+
+  v = gamma * v + matvec(d_history, c_history, v, indices)
+  return v
+
+
+def inv_jacobian_product(pytree: Any,
+                         d_history: Any,
+                         c_history: Any,
+                         gamma: float = 1.0,
+                         start: int = 0):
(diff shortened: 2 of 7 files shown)


zaccharieramzi:
@mblondel the tests are now passing in float32 probably due to some update in main that fixed the issue.

When you have time could you answer the other questions I had regarding this PR?

I think all of them could be handled in different PRs, so I am going to mark this one as ready for review.

jaxopt/_src/broyden.py
```diff
@@ -0,0 +1,382 @@
+# Copyright 2021 Google LLC
```
mblondel, line 1: 2023 :)

jaxopt/_src/broyden.py
```diff
@@ -0,0 +1,382 @@
...
+
+def matvec(d_history, c_history, x, indices):
+  # CD^T x
+  # here j is the history dimension and i is the space dimension
+  dtx = jnp.einsum('j..., ... -> j', d_history[indices], x)
+  cdtx = jnp.einsum('j..., j -> ...', c_history[indices], dtx)
+  return cdtx
+
+def inv_jacobian_product_leaf(v: jnp.ndarray,
+                             d_history: jnp.ndarray,
```
mblondel, line 49: 1 space is missing for the indentation

jaxopt/_src/broyden.py
```diff
@@ -0,0 +1,382 @@
...
+  history_size = len(d_history)
+
+  indices = (start + jnp.arange(history_size)) % history_size
+
+  v = gamma * v + matvec(d_history, c_history, v, indices)
+  return v
+
+
+def inv_jacobian_product(pytree: Any,
+                        d_history: Any,
```
mblondel, line 65: same

jaxopt/_src/broyden.py
```diff
@@ -0,0 +1,382 @@
...
+    verbose: whether to print error on every iteration or not.
+      Warning: verbose=True will automatically disable jit.
+
+  Reference:
+    Charles G. Broyden.
+    A Class of Methods for Solving Nonlinear Simultaneous Equations.
+    Equation (4.5) (page 581).
+  """
+
+  fun: Callable
```
mblondel, line 173: We called it optimality_fun in other root finding classes https://jaxopt.github.io/stable/root_finding.html But now that I think about it, it will probably cause issues if `has_aux=True`...

jaxopt/_src/broyden.py
```diff
@@ -0,0 +1,382 @@
...
+  d_history: Any
+  c_history: Any
+  gamma: jnp.ndarray
+  aux: Optional[Any] = None
+  failed_linesearch: bool = False
+
+
+@dataclass(eq=False)
+class Broyden(base.IterativeSolver):
+  """Broyden solver.
```
mblondel, line 125:
```suggestion
  """(Limited-memory) Broyden solver.
```

jaxopt/_src/broyden.py
```diff
@@ -0,0 +1,382 @@
...
+
```
mblondel, line 126: Could you add here a brief explanation of what are the key differences with LBFGS?

tests/broyden_test.py
```diff
@@ -0,0 +1,149 @@
+# Copyright 2021 Google LLC
```
mblondel, line 1: 2023

mblondel (commented):
Apologies for the very long delay. Thanks a lot for your contribution @zaccharieramzi. I only have some minor comments.

jaxopt/_src/broyden.py
zaccharieramzi: right ofc !

tests/broyden_test.py
zaccharieramzi: yup

jaxopt/_src/broyden.py
zaccharieramzi: yes

jaxopt/_src/broyden.py
zaccharieramzi: corrected

jaxopt/_src/broyden.py
zaccharieramzi: I am not sure I understand the problem with `has_aux=True`. Happy to change to `optimality_fun` anyway.

jaxopt/_src/broyden.py
zaccharieramzi: sure, wdyt now?

mblondel: LGTM, thanks!

jaxopt/_src/broyden.py
mblondel: `optimality_fun` needs to return a pytree of the same shape as the parameter pytree, but if `has_aux=True`, `optimality_fun` will return a tuple `(pytree, aux)`... So keeping the `fun` argument is fine. We'll have to deprecate the `optimality_fun` argument in `Bisection` and others.

mblondel (commented):
Last comment and we're good to merge: can you add the `Broyden` class to `api.rst` and `root_finding.rst`? Thanks!

jaxopt/_src/broyden.py
zaccharieramzi: makes sense!

zaccharieramzi:
@mblondel added : can you tell me what you think of the content of `root_finding.rst` ? not sure exactly what I need in there, but took inspiration from fixed point and unconstrained

docs/root_finding.rst
```diff
@@ -68,3 +68,44 @@ Scipy wrapper
...
+
+    jaxopt.Broyden
+
+Broyden's method is an iterative algorithm suitable for nonlinear root equations in any dimension.
+It is a quasi-Newton method (like L-BFGS), meaning that it uses an approximation of the Jacobian matrix
+at each iteration.
+The approximation is updated at each iteration with a rank-one update.
+This makes the approximation easy to invert using the Sherman-Morrison formula, given it does not use too many
+updates.
+You can control the number of updates with the ``history_size`` argument.
```
mblondel, line 87:
```suggestion
One can control the number of updates with the ``history_size`` argument.
```

docs/root_finding.rst
```diff
@@ -68,3 +68,44 @@ Scipy wrapper
...
+Furthermore, it uses a line search to ensure the rank-one updates are stable.
```
mblondel, line 88:
```suggestion
Furthermore, Broyden's method uses a line search to ensure the rank-one updates are stable.
```

docs/root_finding.rst
```diff
@@ -68,3 +68,44 @@ Scipy wrapper
...
+
+Example::
+
+    from jaxopt import Broyden
+
+    def F(x):
+      return x ** 3 - x - 2
+
+    broyden = Broyden(optimality_fun=F, x0=1.0)
```
mblondel, line 97: fun=F

mblondel: The initialisation should go in `run`. Please ensure that the example runs correctly :)

docs/root_finding.rst
```diff
@@ -68,3 +68,44 @@ Scipy wrapper
...
+    print(broyden.run().params)
+
+
+For implicit differentiation::
+
+    def F(x, factor):
+      return factor * x ** 3 - x - 2
+
+    def root(factor):
+      broyden = Broyden(optimality_fun=F, x0=1.0)
```
mblondel, line 107: Same

docs/root_finding.rst
zaccharieramzi: So sorry about that, they run now

mblondel:
If you can squash, then I'll merge :)

zaccharieramzi:
Squash done

mblondel:
There is an unrelated failure in main. I will merge when CI is back to green.

PR #420, opened 2023-04-05
Fix prox_lasso, prox_elastic_net, to handle pytrees

vroulet:
Fixed prox_lasso and prox_elastic_net to handle pytrees as inputs and floats for hyperparameters.
Currently, the following
```
from jax import numpy as jnp
from jaxopt import prox
x = dict(weight=jnp.ones((16, 16)), bias=jnp.ones(16))
prox.prox_lasso(x)
```
throws the following error: 
```
Traceback (most recent call last):
  File ".../jaxopt/_src/prox.py", line 74, in prox_lasso
    return tree_util.tree_map(fun, x, l1reg)
  File ".../jax/_src/tree_util.py", line 208, in tree_map
    all_leaves = [leaves] + [treedef.flatten_up_to(r) for r in rest]
  File ".../jax/_src/tree_util.py", line 208, in <listcomp>
    all_leaves = [leaves] + [treedef.flatten_up_to(r) for r in rest]
ValueError: Expected dict, got 1.0.
```
Same for `prox_elastic_net`. 
I added a conditional branch depending on the type of the regularization chosen.


jaxopt/_src/prox.py
```diff
@@ -70,8 +70,12 @@ def prox_lasso(x: Any,
   if l1reg is None:
     l1reg = 1.0
 
-  fun = lambda u, v: jnp.sign(u) * jax.nn.relu(jnp.abs(u) - v * scaling)
-  return tree_util.tree_map(fun, x, l1reg)
+  if type(l1reg) == float:
+    fun = lambda u: jnp.sign(u) * jax.nn.relu(jnp.abs(u) - l1reg * scaling)
+    return tree_util.tree_map(fun, x)
```
mblondel, line 75: There is a lot of code duplication. I think it would be more elegant to do
```python
if type(l1reg) == float:
  l1reg = tree_util.tree_ones_like(x) * l1reg
```

jaxopt/_src/prox.py
```diff
@@ -124,9 +128,25 @@ def prox_elastic_net(x: Any,
     hyperparams = (1.0, 1.0)
 
   prox_l1 = lambda u, lam: jnp.sign(u) * jax.nn.relu(jnp.abs(u) - lam)
-  fun = lambda u, lam, gamma: (prox_l1(u, scaling * lam) /
-                               (1.0 + scaling * lam * gamma))
-  return tree_util.tree_map(fun, x, hyperparams[0], hyperparams[1])
+
+  if type(hyperparams[0]) == 'float':
```
mblondel, line 132: And similarly here.

mblondel (commented):
Thank you Vincent!

Could you add non-regression tests? (tests that failed before your changes and now succeed)

jaxopt/_src/prox.py
vroulet: I simplified it.

jaxopt/_src/prox.py
vroulet: Same here.

mblondel (approved):
LGTM!

PR #438, opened 2023-06-07
Convert zoom linesearch as iterative linesearch solver

vroulet:
The goal of this PR is to convert the [zoom_linesearch.py](https://github.com/google/jaxopt/blob/main/jaxopt/_src/zoom_linesearch.py) function into an IterativeLineSearchSolver just as BacktrackingLineSearch or HagerZhangLineSearch. Along the way, some minor bugs have been found, see below, and an attempt to improve readability has been undertaken. Tests have been converted to the new class and are passed.

**Minor bugs report**:
- The length of the interval was not well defined. The bug is detailed in google/jax/issues/16236, I'll wait for confirmation, but I believe it was just a one-line fix. The bug originated from scipy's code I believe.
- The third reference point for the cubic interpolation was not well selected in one case: if high is set to low and low is set to middle, then the reference point should be high and not low as it seems to be the case in the current zoom_linesearch.py procedure lines [201](https://github.com/google/jaxopt/blob/d40b6d736aca6366af6816c9f16498a594850300/jaxopt/_src/zoom_linesearch.py#L201) to [228](https://github.com/google/jaxopt/blob/d40b6d736aca6366af6816c9f16498a594850300/jaxopt/_src/zoom_linesearch.py#L228).
- Compared to HagerZhangLineSearch, rather than initializing the interval in the init_state function, the solver is coded in two stages. The solver switch to the second stage once the interval is found using lax cond. It seems that in the HagerZhangLineSearch, the initialization may return a zero error after init_state which may mean that the linesearch procedure stops after init_state, while returning then the stepsize provided at initialization rather than the stepsize found. I'll investigate that.
- Compared to BacktrackingLineSearch, the value, gradient and descent direction are computed once at initialization and then kept in the state. This means that during the update, the value, gradient, descent direction, and params are not used. This avoids potentially recomputing them at each iteration, but it may not be the expected semantics for iterative linesearch solvers.


vroulet:
I fixed the dtypes, the test passes. However the internal commit d40b6d736aca6366af6816c9f16498a594850300 (unrelated to this PR) is currently breaking the merge on Python 3.7 for all PRs.
Any feedback welcome!

jaxopt/_src/zoom_linesearch2.py
```diff
@@ -0,0 +1,612 @@
+# Copyright 2023 Google LLC
```
mblondel, line 1: It's ok to remove the old code and put your code instead (no need for a new file).

mblondel (commented):
Impressive work. I would plug the new class in all solvers that use the zoom line search, to check that all solver tests still pass. Maybe we can come up with a small utility function so we don't have to duplicate the linesearch instantiation in all solvers...

mblondel:
Can you merge `main`? (the failures have been resolved)

mblondel:
Thanks for the hard work @vroulet! Up to you whether you prefer to start a new PR or not :)

vroulet:
I'll close this PR in favor of the new clean one: https://github.com/google/jaxopt/pull/442.

PR #452, opened 2023-06-28
Adding num function/grad/prox/ls eval in solvers

zaccharieramzi:
For now HZ-LS does not have the right elements in the state so it will fail in all solvers

Added num function/grad/ls eval in the following solvers:
- (L)-BFGS-(B): added num fun/grad/ls
- Anderson: added num fun
- Armijo SGD: for now armijo LS does not return the correct elements so difficult to do, esp. because the loops do not return the number of iterations done
- Bisection: added num fun
- BlockCD: added num fun/prox/grad
- Broyden: added num fun/ls
- FixedPointIteration: added num fun
- Gauss-Newton: not doable, because at some point linear solve does not return the number of iterations done. This comes down to `_cg_solve` returning only the final result.
- (Proximal/Projected)GradientDescent: the fista LS does not return the number of iterations done (same problem as Armijo)
- IterativeRefinement: added num matvec/matvec bar/solve
- LevenbergMarquardt: not sufficient info returned by the solver (+ extremely complex code)
- MirrorDescent: added num fun/grad/proj
- NonLinearCG: added num fun/grad/ls
- OptaxWrapper: I don't think we have this info in every optax state, which in any case can be looked up with the current API
- PolyakSGD: added num fun/grad
- ScipyWrapper(s): added num fun/jac/hess in scipy minimize, added num fun (and num iter) in scipy root

### DIFF
--- a/jaxopt/_src/anderson.py
+++ b/jaxopt/_src/anderson.py
@@ -108,6 +108,8 @@ class AndersonState(NamedTuple):
   residual_gram: jnp.array
   aux: Any
 
+  num_fun_eval: jnp.array = jnp.array(0, base.NUM_EVAL_DTYPE)
+
 
 @dataclass(eq=False)
 class AndersonAcceleration(base.IterativeSolver):
@@ -185,7 +187,8 @@ class AndersonAcceleration(base.IterativeSolver):
                          params_history=params_history,
                          residuals_history=residuals_history,
                          residual_gram=residual_gram,
-                         aux=aux)
+                         aux=aux,
+                         num_fun_eval=jnp.array(1, base.NUM_EVAL_DTYPE))
 
   def update(self,
              params: Any,
@@ -241,7 +244,8 @@ class AndersonAcceleration(base.IterativeSolver):
                                params_history=params_history,
                                residuals_history=residuals_history,
                                residual_gram=residual_gram,
-                               aux=aux)
+                               aux=aux,
+                               num_fun_eval=state.num_fun_eval+1)
 
     return base.OptStep(params=next_params, state=next_state)
 
@@ -253,7 +257,7 @@ class AndersonAcceleration(base.IterativeSolver):
   def __post_init__(self):
     if self.history_size < 2:
       raise ValueError("history_size should be greater or equal to 2.")
-    
+
     if self.has_aux:
       fun_ = self.fixed_point_fun
     else:
--- a/jaxopt/_src/anderson.py
+++ b/jaxopt/_src/anderson.py
@@ -106,9 +106,9 @@ class AndersonState(NamedTuple):
   params_history: Any
   residuals_history: Any
   residual_gram: jnp.array
-  aux: Any
+  aux: Optional[Any] = None
 
-  num_fun_eval: jnp.array = jnp.array(0, base.NUM_EVAL_DTYPE)
+  num_fun_eval: int = 0
 
 
 @dataclass(eq=False)
--- a/jaxopt/_src/armijo_sgd.py
+++ b/jaxopt/_src/armijo_sgd.py
@@ -72,6 +72,9 @@ def armijo_line_search(fun_with_aux, unroll, jit,
     next_params: params after gradient step
     f_next: loss after gradient step
   """
+  # FIXME: (zramzi) this should return the number of iterations,
+  # the number of function and gradient calls to have
+  # these values available down the line.
   next_params = tree_add_scalar_mul(params, -stepsize, grad)
   f_next, _ = fun_with_aux(next_params, *args, **kwargs)
   grad_sqnorm = tree_l2_norm(grad, squared=True)
--- a/jaxopt/_src/base.py
+++ b/jaxopt/_src/base.py
@@ -44,6 +44,7 @@ from jaxopt import tree_util
 AutoOrBoolean = Union[str, bool]
 ArrayPair = Tuple[jnp.ndarray, jnp.ndarray]
 
+NUM_EVAL_DTYPE = 'int32'
 
 class OptStep(NamedTuple):
   params: Any
@@ -55,8 +56,6 @@ class KKTSolution(NamedTuple):
   dual_eq: Optional[Any] = None
   dual_ineq: Optional[Any] = None
 
-
-NUM_EVAL_DTYPE = jnp.int32
 
 # pylint: disable=g-bare-generic
 def _add_aux_to_value_and_grad(value_and_grad: Callable) -> Callable:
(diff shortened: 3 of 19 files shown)


zaccharieramzi:
@mblondel This is ready for review. 
A couple of comments:
- GaussNewton suffers from an inefficiency in function calls -> will tackle in a subsequent PR
- HZ-LS does not have num fun/grad evals
- Backtracking LS for now does not report num grad eval correctly (it is underestimated), but hopefully, the inefficiency will be corrected and this will be consistent
- Some adhoc line searches (armijo, fista) do not report the num of evals because they use `_while_loop` which does not report the number of iterations -> maybe they should be framed with the new paradigm of @vroulet @fabianp

zaccharieramzi:
Actually, I think that adding a unit test at least for num fun eval would be welcome, writing one

zaccharieramzi:
@mblondel this is re-ready for review: the unit test is written and passes, HZ LS now has the num function eval

jaxopt/_src/armijo_sgd.py
```diff
@@ -127,6 +127,9 @@ class ArmijoState(NamedTuple):
   aux: Optional[Any]
   stepsize: float
   velocity: Optional[Any]
+  num_fun_eval: jnp.array = jnp.array(0)
+  num_grad_eval: jnp.array = jnp.array(0)
+  num_line_search_iter: jnp.array = jnp.array(0)
```
mblondel, line 132:
```suggestion
  num_linesearch_iter: jnp.array = jnp.array(0)
```
for consistency

jaxopt/_src/armijo_sgd.py
```diff
@@ -281,12 +287,16 @@ def update(self, params, state, *args, **kwargs) -> base.OptStep:
...
+    new_num_fun_eval = state.num_fun_eval + 1
+    new_num_grad_eval = state.num_grad_eval + 1
 
     goldstein = self.reset_option == 'goldstein'
     stepsize, next_params, f_next = self._armijo_line_search(
       goldstein, self.maxls, params, f_cur, stepsize, grad, self._coef,
       self.decrease_factor, self.increase_factor, self.max_stepsize, args,
       kwargs)
+    # FIXME: update new_num_fun_eval, new_num_grad_eval, new_num_line_search_iter
+    # with outputs from armijo line search
```
mblondel, line 299: Does that mean that the current returned numbers are incorrect?

jaxopt/_src/armijo_sgd.py
mblondel: and same everywhere

tests/common_test.py
```diff
@@ -96,6 +102,60 @@ def fixed_point_fun(params, *args, **kwargs):
...
+    common_kwargs = dict(jit=False, maxiter=30)
+    fpi = jaxopt.FixedPointIteration(fixed_point_fun=fixed_point_fun, **common_kwargs)
+
+    # Gradient solvers and fixed point
+    for solver in (jaxopt.PolyakSGD(fun=fun, **common_kwargs),
+                   fpi,
+                   jaxopt.Broyden(fun=root_fun, **common_kwargs),
+                   jaxopt.AndersonAcceleration(fixed_point_fun=fixed_point_fun, **common_kwargs),
+                   jaxopt.BFGS(fun, **common_kwargs),
+                   jaxopt.LBFGS(fun=fun, **common_kwargs)):
```
mblondel, line 134: Can you add more of them here?

jaxopt/_src/armijo_sgd.py
zaccharieramzi: Yes, this is a problem.

I will remove these values for the returned state, and add an issue specifying which solvers need to be updated to feature these values.

tests/common_test.py
zaccharieramzi: which ones were you thinking about? For the (prox/proj) gradient descent ones because of the current fista line search it is not possible to have the right results.

I added nonlinear CG

zaccharieramzi: (which allowed to correct a zoom LS related bug)

jaxopt/_src/armijo_sgd.py
zaccharieramzi: done

tests/common_test.py
mblondel: All the ones for which you added support for `num_fun_evals`.

zaccharieramzi: Sure : added MirrorDescent, but it's quite a big one

For BlockCD I can't add it because BlockCD does not support non-jitted functions yet (so it's not possible to build the unit test with the global variable like I did). I added a fixme, and will report this in an issue (should be a similar resolution as what @vroulet did with the cond in the LS).

mblondel:
LGTM, apart from squashing.

zaccharieramzi:
I actually need to correct the tests that are now red after the weak type consistency checks. I am planning to force the dtype to be jnp.int32 in all num evals, wdyt?

mblondel:
Yes I think this should be enough
```python
>>> jnp.iinfo(jnp.int32).max
2147483647
```

zaccharieramzi:
@mblondel corrected the dtypes, can you make sure it's ok before I squash the commits?

mblondel:
Is there a rational for `iter_num` being `int64` and `num_grad_eval` int32`?

jaxopt/_src/anderson.py
```diff
@@ -108,6 +108,8 @@ class AndersonState(NamedTuple):
   residual_gram: jnp.array
   aux: Any
 
+  num_fun_eval: jnp.array = jnp.array(0, base.N_EVAL_DTYPE)
```
mblondel, line 111: Sorry, slight preference for `NUM_EVAL_DTYPE` for consistency

zaccharieramzi: sure, fixing rn

jaxopt/_src/base.py
```diff
@@ -56,7 +56,7 @@ class KKTSolution(NamedTuple):
   dual_ineq: Optional[Any] = None
 
 
-N_EVAL_DTYPE = jnp.int64
+NUM_EVAL_DTYPE = jnp.int64
```
mblondel, line 59: Sorry I would prefer if we use int32. Should hopefully be my last comment :)

mblondel: The reason is because I'm worried int64 could be problematic on TPU. Plus, int32 should be enough in most use cases.

mblondel:
Ready for squashing, thanks!

jaxopt/_src/base.py
zaccharieramzi: makes sense, reverted to int64

PR #456, opened 2023-06-29
Add support for has_aux in AndersonAcceleration

### LINKED ISSUE #443: AndersonAcceleration mismatch between code and documentation
The documentation at https://jaxopt.github.io/stable/_autosummary/jaxopt.AndersonAcceleration.html#jaxopt.AndersonAcceleration states in the detailed description of parameters that is has an init value `has_aux` but doesn't in the code and naturally fails when one tries to set has_aux at initialisation.

Would it be possible to add the has_aux flag to the AndersonAcceleration?

Algue-Rythme:
Fix #443

### DIFF
--- a/jaxopt/_src/anderson.py
+++ b/jaxopt/_src/anderson.py
@@ -99,12 +99,14 @@ class AndersonState(NamedTuple):
       fixed_point_fun(params_history) - params_history
     residual_gram: Gram matrix: G.T @ G with G the matrix of residuals
       each column of G is a flattened pytree of residuals_history
+    aux: auxiliary data returned by fixed_point_fun
   """
   iter_num: int
   error: float
   params_history: Any
   residuals_history: Any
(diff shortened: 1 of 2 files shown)


jaxopt/_src/anderson.py
```diff
@@ -128,6 +130,8 @@ class AndersonAcceleration(base.IterativeSolver):
     ridge: ridge regularization in solver.
       Consider increasing this value if the solver returns ``NaN``.
     has_aux: wether fixed_point_fun returns additional data. (default: False)
+      This additional data is not take into account by the fixpoint.
```
mblondel, line 133:
```suggestion
      This additional data is not taken into account by the fixed point.
```

mblondel (commented):
LGTM apart from minor comments below!

Algue-Rythme:
Done!

PR #457, opened 2023-06-29
Add interpolation explanations for Polyak/Armijo API doc

### LINKED ISSUE #324: Specify for which class of problems ArmijoSGD and PolyakSGD are expected to converge
These methods only converge to full precision on over-parameterized problems.

Not specifying this gives the wrong impression that they should always fully converge, which is not true, and can be highly confusing, or worse, give the impression that the implementation is broken

Algue-Rythme:
Fix #324

### DIFF
--- a/jaxopt/_src/armijo_sgd.py
+++ b/jaxopt/_src/armijo_sgd.py
@@ -133,7 +133,10 @@ class ArmijoState(NamedTuple):
 class ArmijoSGD(base.StochasticSolver):
   """SGD with Armijo line search.
 
-  This implementation assumes that the interpolation property holds.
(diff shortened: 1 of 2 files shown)


jaxopt/_src/armijo_sgd.py
```diff
@@ -133,7 +133,10 @@ class ArmijoState(NamedTuple):
 class ArmijoSGD(base.StochasticSolver):
   """SGD with Armijo line search.
 
-  This implementation assumes that the interpolation property holds.
+  This implementation assumes that the "interpolation property" holds (see Vaswani et al):
```
fabianp, line 136:
```suggestion
  This implementation assumes that the "interpolation property" holds, see for example Vaswani et al. 2019 (https://arxiv.org/abs/1905.09997):
```

fabianp (approved):
LGTM with minor changes, thanks!

PR #464, opened 2023-07-04
Fix lbfgsb grad when bounds are specified

### LINKED ISSUE #463: LBFGSB raises TypeError: got an unexpected keyword argument 'bounds' when being differentiated
Simple example (jaxopt 0.7, python 3.11.3, in a jupyter notebook)

```python
import jaxopt
import jax.numpy as jnp
import jax.scipy as jsp
import jax

def pipeline(x, init_pars, bounds, data):
    def fit_objective(pars, data, x):
        return -jsp.stats.norm.logpdf(pars, loc=data+x, scale=1.0)
    solver = jaxopt.LBFGSB(fun=fit_objective, implicit_diff=True, maxiter=500, tol=1e-6)
    return solver.run(init_pars, bounds=bounds, data=data, x=x)[0]

pipeline(0.5, jnp.array(0.0), (jnp.array(0.0), jnp.array(10.0)), 1.0)
# -> Array(1.5, dtype=float32)
```

but if i try to differentiate this wrt x:

```python
jax.grad(pipeline)(0.5, jnp.array(0.0), (jnp.array(0.0), jnp.array(10.0)), 1.0)
```

I get the following stack trace:

```bash
JaxStackTraceBeforeTransformation         Traceback (most recent call last)

File [~/.pyenv/versions/3.11.3/lib/python3.11/asyncio/base_events.py:607](https://file+.vscode-resource.vscode-cdn.net/Users/nsimpson/code/neos/relaxed/~/.pyenv/versions/3.11.3/lib/python3.11/asyncio/base_events.py:607), in run_forever()
    606 while True:
...
   3201             'got an unexpected keyword argument {arg!r}'.format(
   3202                 arg=next(iter(kwargs))))
   3204 return self._bound_arguments_cls(self, arguments)

TypeError: got an unexpected keyword argument 'bounds'
```

phinate:
Resolves #463 by correctly instantiating the function signature, and by adding the `self._grad_fun` alias.

Also adds a test to catch this is working (uses the example from #463).

(I've just signed the google CLA also!)

### DIFF
--- a/jaxopt/_src/lbfgsb.py
+++ b/jaxopt/_src/lbfgsb.py
@@ -21,6 +21,7 @@
 # [2] J. Nocedal and S. Wright.  Numerical Optimization, second edition.
 
 import dataclasses
+import inspect
 import warnings
 from typing import Any, Callable, NamedTuple, Optional, Union
 
@@ -558,6 +559,9 @@ class LBFGSB(base.IterativeSolver):
       params = params.params
     (value, _), grad = self._value_and_grad_with_aux(params, *args, **kwargs)
     return value, grad
+  
+  def _grad_fun(self, params, *args, **kwargs):
+    return self._value_and_grad_fun(params, *args, **kwargs)[1]
 
   def __post_init__(self):
     _, _, self._value_and_grad_with_aux = base._make_funs_with_aux(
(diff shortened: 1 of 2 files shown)


mblondel (approved):
LGTM, thank you!

mblondel:
Merged, thanks! In case you are motivated, I believe [ScipyBoundedMinimize](https://github.com/google/jaxopt/blob/main/jaxopt/_src/scipy_wrappers.py#L398) has the same issue (implicit diff works with positional arguments but not with keyword arguments).

PR #465, opened 2023-07-04
Fix zoom to handle first trial infinite or NaN

vroulet:
If the first trial of the zoom linesearch was NaN, no valid stepsize was found (returned 0 as final stepsize) and the solver would then get stuck. 
Fixed it by switching to zoom into 0 and first stepsize when a NaN or Inf is encountered during the search interval procedure.
Also fix the error to be inf even if it was nan. Otherwise the linesearch stops even if it could have further found a good stepsize.


mblondel (approved):
LGTM, thank you @vroulet

Issue #445, opened 2023-06-22
Optimization of complex parameters


geggo:
When optimizing a real loss function for a complex variable I encountered some issues with jaxopt 0.7. It seems this use case is less widely used. Are the methods also designed for this use case?

1. I had to use the complex conjugate of the gradient. Is this due to the choice for the complex gradient of jax? Should the jaxopt optimizers take this into account?
2. The 'error' (L2 norm of the gradient), which is used for convergence check is calculated as sum of squares. For complex gradients this should be replaced by the sum over the absolute value, or equivalently, the sum of the squared real and imaginary parts. As a simple fix a patched the l-BGFS and GradientDescent methods. I tested that this change also works with real gradients, and when jitted the extra operation is removed for real gradients.

```python
def tree_l2_norm_patched(tree_x, squared=False):
  """Compute the l2 norm ||tree_x||."""
  squared_tree = jaxopt.tree_util.tree_map(lambda x: jnp.square(x.real) + jnp.square(x.imag), tree_x)
  sqnorm = jaxopt.tree_util.tree_sum(squared_tree)
  if squared:
    return sqnorm
  else:
    return jnp.sqrt(sqnorm)

import jaxopt._src
jaxopt._src.lbfgs.tree_l2_norm = tree_l2_norm_patched
jaxopt._src.proximal_gradient.tree_l2_norm = tree_l2_norm_patched
```

Gregor

mblondel:
Thanks a lot for the bug report. Currently, JAXopt hasn't been tested with complex variables, as this use case is less common in ML. There are probably a few places that need to be fixed, both in the `tree_` utils and in the solvers themselves. Pull-requests would be welcome. I suggest with fix the solvers incrementally in several smaller pull requests.

geggo:
Thanks, I started updating the l-BFGS solver and zoom linesearch, see pull request #468

PR #468, opened 2023-07-07
Support for complex parameter optimization

geggo:
I started adding support for complex parameters, beginning with l-BFGS with zoom linesearch, see also issue #445.

The changes are rather straightforward and based on Sorber L, van Barel M, de Lathauwer L, Unconstrained Optimization of real functions in complex variables, SIAM J. Optim, Vol. 22, No. 3, pp. 879-898

* for some data, e.g., stepsize, change dtype to corresponding real dtype
* use complex conjugate of gradient for initial descent direction
* for dot products by vdot(x,y).real
* additionally, for testing, I made changes such that l-BFGS with zero history size falls back to gradient descent with linesearch

For real parameters these changes are removed when jitted, but could introduce some overhead without jitting. 

I plan to add support for more of the unbound optimizers and linesearch methods.

### DIFF
--- a/jaxopt/_src/lbfgs.py
+++ b/jaxopt/_src/lbfgs.py
@@ -26,13 +26,15 @@ from jaxopt._src import base
 from jaxopt._src.linesearch_util import _init_stepsize
 from jaxopt._src.linesearch_util import _setup_linesearch
 from jaxopt._src.tree_util import tree_single_dtype
+from jaxopt._src.tree_util import get_real_dtype
 from jaxopt.tree_util import tree_add_scalar_mul
+from jaxopt.tree_util import tree_conj
 from jaxopt.tree_util import tree_l2_norm
 from jaxopt.tree_util import tree_map
 from jaxopt.tree_util import tree_scalar_mul
 from jaxopt.tree_util import tree_sub
 from jaxopt.tree_util import tree_sum
-from jaxopt.tree_util import tree_vdot
+from jaxopt.tree_util import tree_vdot_real
 
 
 def select_ith_tree(tree_history, i):
@@ -47,7 +49,7 @@ def inv_hessian_product(pytree: Any,
                         start: int = 0):
   """Product between an approximate Hessian inverse and a pytree.
 
-  Histories are pytrees of the same structure as `pytree` except 
+  Histories are pytrees of the same structure as `pytree` except
   that the leaves are arrays of shape `(history_size, ...)`, where
   `...` means the same shape as `pytree`'s leaves.
 
@@ -79,7 +81,7 @@ def inv_hessian_product(pytree: Any,
 
   def body_right(r, i):
     si, yi = select_ith_tree(s_history, i), select_ith_tree(y_history, i)
-    alpha = rho_history[i] * tree_vdot(si, r)
+    alpha = rho_history[i] * tree_vdot_real(si, r)
     r = tree_add_scalar_mul(r, -alpha, yi)
     return r, alpha
   r, alpha = jax.lax.scan(body_right, pytree, indices, reverse=True)
@@ -89,7 +91,7 @@ def inv_hessian_product(pytree: Any,
   def body_left(r, args):
     i, alpha = args
     si, yi = select_ith_tree(s_history, i), select_ith_tree(y_history, i)
-    beta = rho_history[i] * tree_vdot(yi, r)
+    beta = rho_history[i] * tree_vdot_real(yi, r)
     r = tree_add_scalar_mul(r, alpha - beta, si)
     return r, beta
   r, _ = jax.lax.scan(body_left, r, (indices, alpha))
@@ -104,10 +106,10 @@ def compute_gamma(s_history: Any, y_history: Any, last: int):
   # See Numerical Optimization, second edition, equation (7.20).
   # Note that unlike BFGS, the initialization can change on every iteration.
 
-  fun = lambda s_history, y_history: tree_vdot(y_history[last], s_history[last])
+  fun = lambda s_history, y_history: tree_vdot_real(y_history[last], s_history[last])
   num = tree_sum(tree_map(fun, s_history, y_history))
 
-  fun = lambda y_history: tree_vdot(y_history[last], y_history[last])
+  fun = lambda y_history: tree_vdot_real(y_history[last], y_history[last])
   denom = tree_sum(tree_map(fun, y_history))
 
   return jnp.where(denom > 0, num / denom, 1.0)
@@ -144,6 +146,8 @@ class LbfgsState(NamedTuple):
 @dataclass(eq=False)
 class LBFGS(base.IterativeSolver):
   """LBFGS solver.
+
+  Supports complex variables, see second reference.
 
   Attributes:
     fun: a smooth function of the form ``fun(x, *args, **kwargs)``.
@@ -194,10 +198,14 @@ class LBFGS(base.IterativeSolver):
     verbose: whether to print error on every iteration or not.
       Warning: verbose=True will automatically disable jit.
 
-  Reference:
+  References:
     Jorge Nocedal and Stephen Wright.
     Numerical Optimization, second edition.
     Algorithm 7.5 (page 179).
+
+    Laurent Sorber, Marc van Barel, and Lieven de Lathauwer.
+    Unconstrained Optimization of Real Functions in Complex Variables.
+    SIAM J. Optim., Vol. 22, No. 3, pp. 879-898
   """
 
   fun: Callable
@@ -263,21 +271,21 @@ class LBFGS(base.IterativeSolver):
         stepsize=init_params.state.stepsize,
       )
       init_params = init_params.params
-      dtype = tree_single_dtype(init_params)
+      realdtype = get_real_dtype(tree_single_dtype(init_params))
     else:
-      dtype = tree_single_dtype(init_params)
+      realdtype = get_real_dtype(tree_single_dtype(init_params))
       state_kwargs = dict(
           s_history=init_history(init_params, self.history_size),
           y_history=init_history(init_params, self.history_size),
-          rho_history=jnp.zeros(self.history_size, dtype=dtype),
-          gamma=jnp.asarray(1.0, dtype=dtype),
+          rho_history=jnp.zeros(self.history_size, dtype=realdtype),
+          gamma=jnp.asarray(1.0, dtype=realdtype),
           iter_num=jnp.asarray(0),
-          stepsize=jnp.asarray(self.max_stepsize, dtype=dtype),
+          stepsize=jnp.asarray(self.max_stepsize, dtype=realdtype),
       )
     (value, aux), grad = self._value_and_grad_with_aux(init_params, *args, **kwargs)
     return LbfgsState(value=value,
                       grad=grad,
-                      error=jnp.asarray(jnp.inf, dtype=dtype),
+                      error=jnp.asarray(jnp.inf, dtype=realdtype),
                       **state_kwargs,
                       aux=aux,
                       failed_linesearch=jnp.asarray(False),
@@ -304,9 +312,11 @@ class LBFGS(base.IterativeSolver):
     if isinstance(params, base.OptStep):
       params = params.params
 
+    realdtype = state.rho_history.dtype  # avoid recomputation, take realdtype from initialized state
+
     start = state.iter_num % self.history_size
     value, grad = (state.value, state.grad)
-    descent_direction = tree_scalar_mul(-1.0, grad)
+    descent_direction = tree_scalar_mul(-1.0, tree_conj(grad))
     s_history = state.s_history
     y_history = state.y_history
     rho_history = state.rho_history
@@ -364,8 +374,8 @@ class LBFGS(base.IterativeSolver):
       new_num_linesearch_iter = state.num_linesearch_iter
       failed_linesearch = jnp.asarray(False)
     s = tree_sub(new_params, params)
-    y = tree_sub(new_grad, grad)
-    vdot_sy = tree_vdot(s, y)
+    y = tree_conj(tree_sub(new_grad, grad))
+    vdot_sy = tree_vdot_real(s, y)
     rho = jnp.where(vdot_sy == 0, 0, 1. / vdot_sy)
 
     if self.history_size:
@@ -376,15 +386,14 @@ class LBFGS(base.IterativeSolver):
     if self.history_size and self.use_gamma:
       gamma = compute_gamma(s_history, y_history, start)
     else:
-      gamma = jnp.array(1.0)
+      gamma = jnp.array(1.0, dtype=realdtype)
 
-    dtype = tree_single_dtype(params)
     error = tree_l2_norm(new_grad)
     new_state = LbfgsState(iter_num=state.iter_num + 1,
                            value=new_value,
                            grad=new_grad,
-                           stepsize=jnp.asarray(new_stepsize),
-                           error=jnp.asarray(error, dtype=dtype),
+                           stepsize=jnp.asarray(new_stepsize, dtype=realdtype),
+                           error=jnp.asarray(error, dtype=realdtype),
                            s_history=s_history,
                            y_history=y_history,
                            rho_history=rho_history,
--- a/jaxopt/_src/tree_util.py
+++ b/jaxopt/_src/tree_util.py
@@ -104,6 +104,23 @@ def tree_vdot(tree_x, tree_y):
   return tree_reduce(operator.add, vdots)
 
 
+def _vdot_real(x, y):
+  """Vector dot-product guaranteed to have a real valued result despite
+     possibly complex input. Thus neglects the real-imaginary cross-terms.
+     The result is a real float.
+  """
+  #result = _vdot(x.real, y.real)
+  #if jnp.iscomplexobj(x) and jnp.iscomplexobj(y):
+  #  result += _vdot(x.imag, y.imag)
+  result = _vdot(x, y).real  # NOTE: without jit this is faster than variant above, no difference with jit
+  return result
+
+
+def tree_vdot_real(tree_x, tree_y):
+  """Compute the real part of the inner product <tree_x, tree_y>."""
+  return sum(tree_leaves(tree_map(_vdot_real, tree_x, tree_y)))
+
+
 def tree_dot(tree_x, tree_y):
   """Compute leaves-wise dot product between pytree of arrays.
 
@@ -120,7 +137,7 @@ def tree_sum(tree_x):
 
 def tree_l2_norm(tree_x, squared=False):
   """Compute the l2 norm ||tree_x||."""
-  squared_tree = tree_map(jnp.square, tree_x)
+  squared_tree = tree_map(lambda leaf: jnp.square(leaf.real) + jnp.square(leaf.imag), tree_x)
   sqnorm = tree_sum(squared_tree)
   if squared:
     return sqnorm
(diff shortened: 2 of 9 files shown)


mblondel:
Thanks for starting this! It looks like your branch is not up to date. You need to merge `main` in your branch first.

jaxopt/_src/lbfgs.py
```diff
@@ -53,7 +54,7 @@ def inv_hessian_product_leaf(v: jnp.ndarray,
   indices = (start + jnp.arange(history_size)) % history_size
 
   def body_right(r, i):
-    alpha = rho_history[i] * jnp.vdot(s_history[i], r)
+    alpha = rho_history[i] * jnp.vdot(s_history[i], r).real
```
vroulet, line 52: Just to be sure that I understand:
- in theory the dot product should be real because the function is real and one rounds it for numerical stability or does the algorithm really require taking the real part?

jaxopt/_src/zoom_linesearch.py
```diff
@@ -633,9 +636,9 @@ def init_state(
       _, aux = self._fun_with_aux(params, *args, **kwargs)
 
     if descent_direction is None:
-      descent_direction = tree_scalar_mul(-1.0, grad)
+      descent_direction = tree_scalar_mul(-1.0, tree_conj(grad))
 
-    slope = tree_vdot(grad, descent_direction)
+    slope = tree_vdot(grad, descent_direction).real
```
vroulet, line 642: Same as for lbfgs: 
Just to be sure that I understand:
- in theory the dot product should be real because the function is real and one rounds it for numerical stability or does the algorithm really require taking the real part?

I ask because in e.g. the implementation of linear solvers in complex values similar issues arose and different patches for numerical stability where used. 
Once we have a test on a simple function we may select what's the best if the problem is in terms of numerical stability

vroulet (commented):
Hello Gregor! 

Thanks for starting this! I've left you some comments. Happy to discuss that more.

jaxopt/_src/lbfgs.py
geggo: The algorithm indeed requires to take the real part. Then it is identical to the situation when the complex arguments are split into real and imaginary parts. `jnp.vdot(x, y).real` could also be expressed as `jnp.vdot(x.real, y.real) + jnp.vdot(x.imag, y.imag)`, which might be more efficient, need to benchmark it. For the moment I chose the simpler expression.

jaxopt/_src/zoom_linesearch.py
geggo: ...same as above

geggo:
Thanks for the comments, I will try to incorporate them all, starting with adding tests.

jaxopt/_src/lbfgs.py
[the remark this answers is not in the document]
geggo: added reference 

Laurent Sorber, Marc van Barel, and Lieven de Lathauwer.
Unconstrained Optimization of Real Functions in Complex Variables.
SIAM J. Optim., Vol. 22, No. 3, pp. 879-898

to doc string of LBFGS class.

jaxopt/_src/tree_util.py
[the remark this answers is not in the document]
geggo: Done.
* Use complex arrays for testing of tree_util
* fix test for tree_l2_norm to support complex arrays
* add test for tree_conj

jaxopt/_src/lbfgs.py
vroulet: Great, we may add a function vdot_real in tree_utils like the one already present in jax (see https://github.com/google/jax/blob/1795b12a9f7bca118d75c981fadd61331680de57/jax/_src/scipy/sparse/linalg.py#L39). Using the second option (`jnp.vdot(x.real, y.real) + jnp.vdot(x.imag, y.imag)`) is a priori slightly faster I think.

vroulet:
Hello Gregor, 

Thanks for the changes! As I told you, there was some revamping necessary in tree_util.
In particular the `tree_single_dtype` was not robust yet (independent of your PR).
I'd propose you use a function `get_real_dtype`  and `tree_vdot_real` mostly for readability reasons and use it then in your modifications of `lbfgs.py` and `zoom_linesearch.py`. No need to make get_real_dtype public, but making tree_vdot_real public would make sense (that is adding it to the jaxopt.tree_util.py imports). Also one may add tree_real and tree_imag in addition to tree_conj in case it's needed later on (all public)

I prepared the changes: 
- could you make the following changes in `_src/tree_util.py`?
```
import functools
import jax
from jax import numpy as jnp
import numpy as onp
from jax import tree_util as tu

# Already present in _src/tree_util.py
_vdot = functools.partial(jnp.vdot, precision=jax.lax.Precision.HIGHEST)
tree_map = tu.tree_map

# New or modified in _src/tree_util.py
tree_real = functools.partial(tree_map, jnp.real)
tree_real.__doc__ = "Real part of a tree."

tree_imag = functools.partial(tree_map, jnp.imag)
tree_imag.__doc__ = "Imaginary part of a tree."

tree_conj = functools.partial(tree_map, jnp.conj)
tree_conj.__doc__ = "Complex conjugate of a tree."


def get_real_dtype(dtype):
    """Dtype corresponding of real part of a complex dtype."""
    if dtype not in [f'complex{i}' for i in [4, 8, 16, 32, 64, 128]]:
        return dtype
    else:
        return dtype.type(0).real.dtype


def tree_single_dtype(tree, convert_in_jax_dtype=True):
    """The dtype for all values in a tree, provided that all leaves share the same type.

    If the leaves have different type, raise a ValueError.

    Args:
      tree: tree to get the dtype of
      convert_in_jax_type: whether to convert the types in JAX precision.
        Namely, a numpy int64 type is converted in a jax.numpy int32 type
        by default unless one enables double precision using
        jax.config.update("jax_enable_x64", True)
    Return:
      dtype shared by all leaves of the tree
    """
    if convert_in_jax_dtype:
        dtypes = set(
            jnp.asarray(p).dtype
            for p in tu.tree_leaves(tree)
            if isinstance(
                p, (bool, int, float, complex, onp.ndarray, jnp.ndarray)
            )
        )
    else:
        dtypes = set(
            onp.asarray(p).dtype
            for p in tu.tree_leaves(tree)
            if isinstance(
                p, (bool, int, float, complex, onp.ndarray, jnp.ndarray)
            )
        )
    if not dtypes:
        return None
    if len(dtypes) == 1:
        dtype = dtypes.pop()
        return dtype
    raise ValueError("Found more than one dtype in the tree.")

def _vdot_real(x, y):
  """Vector dot-product guaranteed to have a real valued result despite
     possibly complex input. Thus neglects the real-imaginary cross-terms.
     The result is a real float.
  """
  result = _vdot(x.real, y.real)
  if jnp.iscomplexobj(x) or jnp.iscomplexobj(y):
    result += _vdot(x.imag, y.imag)
  return result


def tree_vdot_real(x, y):
  """Compute the real part of the inner product <tree_x, tree_y>."""
  return sum(tu.tree_leaves(tree_map(_vdot_real, x, y)))
```
- could you also add the following tests in `tests/tree_util_test.py`?
```
from absl.testing import absltest
from absl.testing import parameterized

import jax
import jax.numpy as jnp
import jax.tree_util as jtu

from jaxopt._src import test_util

import numpy as onp


class TreeUtilTest(test_util.JaxoptTestCase):
  def setUp(self):
    super().setUp()
    rng = onp.random.RandomState(0)

    self.tree_A = (rng.randn(20, 10) + 1j*rng.randn(20, 10), rng.randn(20))
    self.tree_B = (rng.randn(20, 10), rng.randn(20))

    self.tree_A_dict = (1.0, {"k1": 1.0, "k2": (1.0, 1.0)}, 1.0)
    self.tree_B_dict = (1.0, {"k1": 2.0, "k2": (3.0, 4.0)}, 5.0)

    self.array_A = rng.randn(20) + 1j*rng.randn(20)
    self.array_B = rng.randn(20)

  def test_get_real_dtype(self):
    complex_dtype = jnp.asarray(1j*1.).dtype
    real_dtype = jnp.asarray(1.).dtype
    self.assertEqual(get_real_dtype(complex_dtype), real_dtype)

  def test_tree_conj(self):
    expected = onp.conj(self.array_A)
    got = tree_conj(self.array_A)
    self.assertAllClose(expected, got)

    expected = (onp.conj(self.tree_A[0]), onp.conj(self.tree_A[1]))
    got = tree_conj(self.tree_A)
    self.assertAllClose(expected, got)

  def test_tree_imag(self):
    expected = onp.imag(self.array_A)
    got = tree_imag(self.array_A)
    self.assertAllClose(expected, got)

    expected = (onp.imag(self.tree_A[0]), onp.imag(self.tree_A[1]))
    got = tree_imag(self.tree_A)
    self.assertAllClose(expected, got)

  def test_tree_real(self):
    expected = onp.real(self.array_A)
    got = tree_real(self.array_A)
    self.assertAllClose(expected, got)

    expected = (onp.real(self.tree_A[0]), onp.real(self.tree_A[1]))
    got = tree_real(self.tree_A)
    self.assertAllClose(expected, got)

  @parameterized.product(high_precision=[True, False], convert_in_jax_dtype=[True, False])
  def test_tree_single_dtype(self, high_precision, convert_in_jax_dtype):
    current_precision = jax.config.jax_enable_x64
    if high_precision:
      jax.config.update("jax_enable_x64", True)
    else:
      jax.config.update("jax_enable_x64", False)

    dtype_conversions = {        
      onp.asarray(True).dtype: jnp.asarray(True).dtype,
      onp.asarray(0).dtype: jnp.asarray(0).dtype,
      onp.asarray(0.0).dtype: jnp.asarray(0.0).dtype,
      onp.asarray(1j * 1.0).dtype: jnp.asarray(1j * 1.0).dtype,
    }

    test_base_types = [
      (True, False),
      (1, 2),
      (3.0, 4.0),
      (1.0 + 1j * 2.0, 2.0 + 1j * 3.0),
      (onp.asarray((1.0, 3.0)), onp.asarray((2.0, 4.0))),
      (jnp.asarray((1.0, 2.0)), jnp.asarray((3.0, 4.0))),
    ]
    for x in test_base_types:
      dtype_tree = tree_single_dtype(
        x, convert_in_jax_dtype=convert_in_jax_dtype
      )
      dtype_leaf = onp.asarray(x[0]).dtype
      if not high_precision and convert_in_jax_dtype:
        if dtype_leaf not in (onp.asarray(True).dtype, jnp.asarray(0.0).dtype):
          self.assertNotEqual(dtype_tree, dtype_leaf)
          dtype_leaf = dtype_conversions[dtype_leaf]
          self.assertEqual(dtype_tree, dtype_leaf)
        else:
          self.assertEqual(dtype_tree, dtype_leaf)
      else:
        self.assertEqual(dtype_tree, dtype_leaf)

    self.assertRaises(ValueError, tree_single_dtype, (1., 1))
    class MyInt():
      i: int = 1
    self.assertEqual(None, tree_single_dtype(MyInt()))

    jax.config.update("jax_enable_x64", current_precision)

  def test_tree_vdot_real(self):
    expected = onp.vdot(self.array_A, self.array_B).real
    got = tree_vdot_real(self.array_A, self.array_B)
    self.assertAllClose(expected, got)

    expected = sum([onp.vdot(self.tree_A[i], self.tree_B[i]).real for i in range(2)])
    got = tree_vdot_real(self.tree_A, self.tree_B)
    self.assertAllClose(expected, got)

if __name__ == "__main__":
  absltest.main()
```

Ping me when you have a test ready for lbfgs in complex values. Again I'm happy to help if you get unreadable errors.

jaxopt/_src/lbfgs.py
geggo: Some benchmarking, see
[https://gist.github.com/geggo/d1a990b7d7c6ce97180dbb00c15f1ca0]
shows that `jnp.vdot(x, y).real`  is actually faster than `jnp.vdot(x.real, y.real) + jnp.vdot(x.imag, y.imag)` when non jitted, no difference when jitted.

Furthermore, I checked that only if both arrays are complex, the second term of `jnp.vdot(x.real, y.real) + jnp.vdot(x.imag, y.imag)` is needed because it is zero if either x or y real arrays. It seems the jax implementation could be improved - or I am missing something?

geggo:
Thanks for the detailed help, I made the changes as suggested, tests are ok, but I did not yet update the dtype handling. Still work in progress.

For testing I use a jupyter notebook (https://gist.github.com/geggo/1635d11cd5742205b7229f3b65d5cecb) , results for function of complex variables now gives exactly the same result as the equivalent problem for real variables. Will try to convert this to a test case.

One issue I found that the routine for the inverse hessian product in lbgfs.py gives different results if real and imaginary parts are stored in different leafs of a pytree, or as e.g. two columns of an array, respectively as real and imaginary part of a complex array.

I think the loop should be expressed in terms of tree_XXX operations. Interestingly, the current implementation performs well for the test problem, although it should not. Have to look further into this.

Thanks for all your support
Gregor

vroulet:
Hello Gregor,

Thanks for the notebook! I don't fully understand the bug you mention. Could you make a minimal reproducing code for me to understand? Thanks!

vroulet:
Hello Gregor,

Thank you! That's a very good catch. The current implementation with pytrees is indeed incorrect. You summarized it well.
For the record (other readers), mathematically, for a vector a = (a1, a2) and a rank one matrix u v^T with u = (u1, u2), v = (v1, v2), the matrix vector product reads a priori
`uv^T a = <v, a> u = ((a1*v1 + a2v2) u1, (a1v1 + a2v2)u2)`
The current implementation with pytrees compute in its loops 
`(a_1v_1 u_1, a2v2 u2)`
which is quite different.
Below is a fix that passes your test. Could you 
1. modify accordingly `lbfgs.py`, 
2. add the test you had to the `lbfgs_test.py`,
3. make a test to see if now the whole lbfgs routine returns the same result after, say 5 iterations, whether it is a pytree or the matrix version (just as you did but for the update function now). 

Thank you again!

```
from typing import Any
import jax
import jax.numpy as jnp
from jax.tree_util import tree_map
from jaxopt.tree_util import tree_vdot_real, tree_add_scalar_mul

def select_ith_tree(tree, i):
  return tree_map(lambda a: a[i, ...], tree)

def inv_hessian_product(v: Any,
                        s_history: Any,
                        y_history: Any,
                        rho_history: jnp.ndarray,
                        gamma: float = 1.0,
                        start: int = 0):
  """Product between an approximate Hessian inverse and a pytree.

  Histories are pytrees of the same structure as `pytree` execept 
  that the leaves are arrays of shape `(history_size, ...)`, where
  `...` means the same shape as `pytree`'s leaves.

  The notation follows the reference below.

  Args:
    pytree: pytree to multiply with.
    s_history: pytree whose leaves contain parameter residuals,
      i.e., `s[k] = x[k+1] - x[k]`
    y_history: pytree whose leaves contain gradient residuals,
      i.e., `y[k] = g[k+1] - g[k]`.
    rho_history: array containing `rho[k] = 1. / vdot(s[k], y[k])`.
    gamma: scalar to use for the initial inverse Hessian approximation,
      i.e., `gamma * I`.
    start: starting index in the circular buffer.

  Returns:
    Product between approximate Hessian inverse and the pytree

  Reference:
    Jorge Nocedal and Stephen Wright.
    Numerical Optimization, second edition.
    Algorithm 7.4 (page 178).
  """

  history_size = rho_history.shape[0]

  indices = (start + jnp.arange(history_size)) % history_size

  def body_right(r, i):
    si, yi = select_ith_tree(s_history, i), select_ith_tree(y_history, i)
    alpha = rho_history[i] * tree_vdot_real(si, r)
    r = tree_add_scalar_mul(r, -alpha, yi)
    return r, alpha
  r, alpha = jax.lax.scan(body_right, v, indices, reverse=True)

  def body_left(r, args):
    i, alpha = args
    si, yi = select_ith_tree(s_history, i), select_ith_tree(y_history, i)
    beta = rho_history[i] * tree_vdot_real(yi, r)
    r = tree_add_scalar_mul(r, alpha - beta, si)
    return r, beta
  r, _ = jax.lax.scan(body_left, r, (indices, alpha))

  r = r * gamma

  return r

def test():
  s_history = jnp.arange(24).reshape(4,2,3)
  y_history = jnp.ones((4,2,3))
  rho_history = jnp.ones(4)

  s_history0 = s_history[:,0,:]
  s_history1 = s_history[:,1,:]

  y_history0 = y_history[:,0,:]
  y_history1 = y_history[:,1,:]

  v = jnp.ones((2,3))
  v0 = v[0]
  v1 = v[1]
  a = inv_hessian_product(v, s_history, y_history, rho_history, 1, 0)
  b = inv_hessian_product((v0, v1), (s_history0, s_history1), (y_history0, y_history1), rho_history, 1, 0)
  b = jnp.stack((b[0], b[1]))
  assert jnp.allclose(a, b)
  print('Test passed')

if __name__ == '__main__':
  test()
```

mblondel:
@vroulet Can you fix it separately? Then @geggo can merge your PR.

geggo:
I merged PR #475 with modifications for complex parameters (and also main). 
I checked that now complex optimization gives the same results as with the corresponding real case (and now both for real and imaginary part stored in an array  or in pytree), for zoom, backtracking, and Hager-Zhang linesearch.

I also adapted the Nonlinear conjugate gradient method to support complex gradients.

tests/lbfgs_test.py
```diff
@@ -358,7 +358,45 @@ def _run_lbfgs_solver(X, y, lmbd, n_iter):
...
+    def loss_complex(z):
+      return jnp.sum(jnp.abs(f(z)) ** 1.5)
+
+    def loss_real(zR):
+      return loss_complex(R2C(zR))
+
+    z0 = jnp.array([1 - 1j, 0 + 1j])
+    xy0 = jnp.stack((z0.real, z0.imag))
+
+    solver_C = LBFGS(fun=loss_complex, maxiter=5, history_size=5, stepsize=-1)
```
vroulet, line 454: Add linesearch argument

jaxopt/_src/lbfgs.py
```diff
@@ -202,10 +202,14 @@ class LBFGS(base.IterativeSolver):
     verbose: whether to print error on every iteration or not.
       Warning: verbose=True will automatically disable jit.
 
-  Reference:
+  References:
```
vroulet, line 195: Add at start of doc:
``Supports complex variables, see second ref.''

jaxopt/_src/lbfgs.py
```diff
@@ -318,11 +319,14 @@ def update(self,
     start = state.iter_num % self.history_size
     value, grad = (state.value, state.grad)
 
-    product = inv_hessian_product(pytree=grad, s_history=state.s_history,
+    if self.history_size:
```
vroulet, line 312: Could you explain why we need to fall back like that? At least give a comment with the reason for the change please

vroulet:
Hello Gregor!

Thanks for the hard work! The test seems good (just add the linesearch argument).  
For the nonlinear conjugate gradient, it would be better to do it in a separate PR. Could you stash somewhere the changes you have for the nonlinear conjugate gradient such that we'll review them later in a separate PR?
I left you additional comments. 
Could you add tests for the support of complex numbers for the linesearch separately? (otherwise if these files are changed we may not see where the bugs come from). Also add in the docs of the linesearches that they can handle complex numbers and point to the ref. 
Once the PR #475 is merged, you'll need to rebase your code with the main branch.

tests/lbfgs_test.py
geggo: Thanks, fixed.

jaxopt/_src/lbfgs.py
geggo: inv_hessian_product fails for zero history_length. Moved test to inv_hessian.

jaxopt/_src/lbfgs.py
geggo: Done

geggo:
Hi Vincent,

thanks for the suggestions, tried to implement them all. Adding tests for linesearch with complex variables (in new file) was a good idea, revealed a bug when descent direction is not given.

jaxopt/_src/lbfgs.py
vroulet: Ok. But was this specific for the complex case or already an issue before? Would you have a test case that was resulting in a bug? Also can't it be handled by the fact that a loop over an empty set just returns the initial value?

mblondel:
Looks like you need to merge `main` once again. Thanks for this huge contribution @geggo!

jaxopt/_src/lbfgs.py
geggo: If I remember correctly, this was already an issue before changing the inverse Hessian loop, LBFGS with history_size=0 gave an error, but possibly at different place.

It seems that when tracing the function the loop is always entered to determine shape, but fails when indexing zero length axes. 

The shortened traceback that I get is:

File ~/Devel/jaxopt/jaxopt/_src/lbfgs.py:41, in select_ith_tree(tree_history, i)
     39 def select_ith_tree(tree_history, i):
     40   """Select tree corresponding to ith history entry"""
---> 41   return tree_map(lambda a: a[i, ...], tree_history)

[...]

IndexError: index is out of bounds for axis 0 with size 0

With this check for history_size==0, LBFG falls back to steepest descent with linesearch, but without scaling of linestep by gamma. It might be useful to keep this gamma scaling, because then stepsize=1 is always (?) a good starting guess. But this would need to always keep one level of history.

vroulet: Ok, great! You're finding many bugs, thank you very much! I'll investigate that in a separate PR like the bug you found for the pytrees. Again, thank you Gregor!

vroulet:
Hello Gregor, 
I made a separate PR to handle `history_size = 0` with associated test, see #482 .
Once it's merged (check maybe in two days), you can merge with main (another PR will also have been merged in the meantime). Then squash everything and I think we'll be good to go!

geggo:
Thanks for all your help, I merged main branch, my tests are ok. Squashing all the commits is something I need to do?

vroulet:
Hello Gregor,
Thank you for all the work!
We prefer to do one PR=one commit. If you encounter any difficulty, just tell me, I should be able to squash from your fork if necessary.

geggo:
I would appreciate if you could do the squashing when merging - I do not know how to accomplish this on my branch. Thanks for your help!

vroulet (approved):
LGTM, thanks again Gregor! Everything has been squashed and will be merged.

Issue #409, opened 2023-03-03
jaxopt returns the initial values as the final solution for binary logit model


PepSalehi:
I am trying to perform a maximum likelihood estimation of the binary logit model. The sample code generates a sample of data using the true parameters (`[0.5,0.5]`). I then use `scipy`, `jax.scipy`, and `jaxopt` to estimate the parameters. 
As you can see, `scipy` returns the closest result, while `jax.scipy` is somewhat decent. But no matter the initial value, `jaxopt` always returns that initial value as the solution:

```
import jax
import jax.numpy as jnp
print("Using jax", jax.__version__)
import jaxopt 
import jax.scipy.optimize as jsp_opt
import numpy as np 
import scipy.optimize as opt

rng = jax.random.PRNGKey(42)
N = 1000
beta = jnp.array([[0.5,0.5]]).T
income = jax.random.normal(rng, shape=(N,1))
x = jnp.hstack([jnp.ones((N,1)), income])

def simulate_binary_logit(x, beta):
    """
    PURPOSE: generate binary outcomes according to
 % the logit specification.
    INPUTS: x : NxJ vector of independent variable
 % Beta : Jx1 parameter vector
 %------------------------------------------------
 % OUTPUT: y : N-by-1 binary choice vector
 %------------------------------------------------
    """
    eps = 0.01
    beta = beta.reshape(-1,1)
    N = x.shape[0]
    J = beta.shape[0]

    epsilon = jax.random.gumbel(rng,shape =(N,J))
    Beta_augmented = jnp.hstack([beta, jnp.zeros_like(beta)])
    utility = x @ Beta_augmented + epsilon
    
    choice_idx = np.argmax(utility, axis=1)
    return (choice_idx).reshape(-1,1)

y = simulate_binary_logit(x, beta)
# numpy version
def binary_logit_log_likelihood(beta, y,x):
    beta = beta.reshape(-1,1)
    lambda_xb = np.exp(x@beta) / (1 + np.exp(x@beta))
    ll_i = y * np.log(lambda_xb) + (1-y) * np.log(1-lambda_xb)
    ll = -np.sum(ll_i)
    return ll

# jax version 
def binary_logit_log_likelihood_jax(beta, y,x):
    # beta = beta.reshape(-1,1)
    lambda_xb = jnp.exp(x@beta) / (1 + jnp.exp(x@beta))
    ll_i = y * jnp.log(lambda_xb) + (1-y) * jnp.log(1-lambda_xb)
[...]

PR #473, opened 2023-07-12
Fixed bug 409 due to bad handling of NaN values in zoom linesearch

vroulet:
Fixed bug [409](https://github.com/google/jaxopt/issues/409), added corresponding test, enhanced debugging messages (for linesearch and solvers).

### DIFF
--- a/jaxopt/_src/zoom_linesearch.py
+++ b/jaxopt/_src/zoom_linesearch.py
@@ -34,6 +34,13 @@ from jaxopt.tree_util import tree_vdot
 # pylint: disable=g-bare-generic
 # pylint: disable=invalid-name
 
+_FLAG_NOT_DESCENT_DIRECTION = 1
+_FLAG_CURVATURE_COND_NOT_SATSIFIED = 2
+_FLAG_MAX_ITER_REACHED = 4
+_FLAG_INTERVAL_TOO_SMALL = 8
+_FLAG_NAN_INF_VALUES = 16
+_FLAG_NO_STEPSIZE_FOUND = 32
+
 _dot = functools.partial(jnp.dot, precision=lax.Precision.HIGHEST)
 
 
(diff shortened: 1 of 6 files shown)


tests/lbfgs_test.py
```diff
@@ -358,6 +359,56 @@ def _run_lbfgs_solver(X, y, lmbd, n_iter):
...
+
+    y = simulate_binary_logit(x, beta)
+    y = jnp.ravel(y)
+
+    # numpy version
+    def binary_logit_log_likelihood(beta, y,x):
+      lambda_xb = onp.exp(x@beta) / (1 + onp.exp(x@beta))
+      ll_i = y * onp.log(lambda_xb) + (1-y) * onp.log(1-lambda_xb)
+      ll = -onp.sum(ll_i)
+      return ll
```
mblondel, line 391: remove

tests/lbfgs_test.py
```diff
@@ -358,6 +359,56 @@ def _run_lbfgs_solver(X, y, lmbd, n_iter):
...
+      ll_i = y * jnp.log(lambda_xb) + (1-y) * jnp.log(1-lambda_xb)
+      ll = -jnp.sum(ll_i)
+      return ll
+
+    beta_init = jnp.array([0.01,0.01])
+
+    # using scipy
+    scipy_res = opt.minimize(fun=binary_logit_log_likelihood, 
+                             args=(onp.asarray(y),onp.asarray(x)), 
+                             x0 = (onp.asarray(beta_init)), method='BFGS')
```
mblondel, line 405: remove

jaxopt/_src/zoom_linesearch.py
```diff
@@ -106,17 +106,17 @@ def _set_val(x, y):
 
 def _check_failure_status(fail_code):
   """Print failure reason according to fail value."""
-  if fail_code == 1:
+  if fail_code % 2 == 0:
```
mblondel, line 109: Using bitwise flags would be more standard (I think).

tests/lbfgs_test.py
vroulet: Which part do you want me to remove exactly? Having the comparison with scipy looks like a good safeguard. By the way I think we can remove the test test_first_stepsize_inf which did not reveal all issues.

mblondel: I thought it's sufficient to just check that we don't get stuck at zero (our test suite is getting longer and longer to run). But you're right, maybe it's better to compare with scipy to be entirely sure.

vroulet: Ok, I removed the previous test as it was redundant and less documented and kept the comparison to scipy.

vroulet:
I've also modified the way the linesearch would handle the case when the max_stepsize was reached. Before, once the maximal stepsize was reached the algorithm would keep on trying to increase it to find an interval for which the curvature error would be satisfied, while always staying to the max_stepsize. So there would be a waste of function/gradient evaluations. I fixed this by letting the linesearch stop if the max_stepsize is reached and no interval was found (meaning that the sufficient decrease condition is satisfied but not the curvature one and the latter would not be satisfied anyway). I documented it in the code.

jaxopt/_src/zoom_linesearch.py
```diff
@@ -138,7 +150,7 @@ class ZoomLineSearchState(NamedTuple):
 
   error: float
   done: bool
-  fail_code: int  # encode failure status, see _check_status
+  error_flag: int  # encode failure status, see _check_status
```
mblondel, line 153: What's the motivation for changing the name? Since, other solvers use `fail_code`, this would create an inconsistency.

vroulet: The NaN/Inf flag does not necessarily lead to a failure but was more of a warning. One could a priori fully remove it now too. Other than that, we can use fail_code. 
PS: Which other algorithms use fail_code? I could further harmonize if you want.

mblondel: But in that case error != warning too. `fail_code` is only used in the zoom linesearch but `failed_linesearch` is used in many places so I would prefer to keep `fail_code` for consistency.

vroulet: Ok! The changes have been made, I've squashed everything.

Issue #476, opened 2023-07-13
Add typing-extensions dependency to pip package


Atomotron:
On Python 3.11.2, the typing-extensions package has to be installed manually for jaxlib to work. Here is the file that imports it: https://github.com/google/jaxopt/blob/3a941ef07e22188b208a15a08bd9bff2f2264207/jaxopt/_src/levenberg_marquardt.py#L24 

I don't see where the package is defined, so I can't reference the list of dependencies that it's missing from.

mblondel:
fixed by #479

PR #479, opened 2023-07-17
Adding tests in python 3.11, fixing typing extension issue

vroulet:
Current workflow is not tested on python 3.11, which generated issues, see #476. This PR attempts to fix the typing issue and add tests on python 3.11.


mblondel (approved):
LGTM, ty!

PR #482, opened 2023-07-19
Let lbfgs handle zero history size

vroulet:
In that case LBFGS is just Gradient descent, added test. 
PS: removed line `last = (start + self.history_size) % self.history_size` since it was necessarily equal to start and was a bit confusing (the circular buffer is neat :) ).


mblondel (approved):
LGTM, thank you

PR #490, opened 2023-08-02
Improvements to the resnet cifar10 example

fabianp:
1. Reaches 91% accuracy with a resnet18 model on CIFAR10, before
we were only getting 88%. The main change that made this possible was
to tweak the size of the convolutional kernel in the first layer, from
7x7 (default) to 3x3. This is documented in the code, which now allows
easily to overwrite the default kernel size.

2. Runs in 50 epochs (vs 200 before), and only takes 13 min on free
colab, vs 1 hour before.

3. Removed many unused variables, and cleaned up the code.

4. I'm now using a warm-up learning rate, which seems to be working
great across different architectures and datasets.


docs/notebooks/deep_learning/resnet_flax.md
```diff
@@ -130,76 +181,73 @@ def augment(image, label):
...
 
-With the data ready, we can now define the model. Below we define the ResNet architecture that we'll later instantiate. We define different variants of the architecture with different sizes and depths (`'ResNet1'`, `'ResNet18'`, `'ResNet34'`, `'ResNet50'` and `'ResNet101'`). The architecture is based on the [Flax imagenet example](https://github.com/google/flax/blob/main/examples/imagenet/models.py).
+The following code is based on the [Flax imagenet example](https://github.com/google/flax/blob/main/examples/imagenet/models.py). The only difference with that code is the addition of the keyword argument `initial_conv_config` to the `ResNet` class, which allows to change the configuration of the initial convolutional layer of the network. This is important to get state of the art accuracy on CIFAR10, as the default kernel size (7, 7) is too big for the small 32x32 images of CIFAR10.
 
 ```{code-cell} ipython3
-:id: onkVLRw7L3j4
+:id: P7Z3Vex8QuGz
+
+ModuleDef = Any
```
mblondel, line 212: What do we gain from this definition? It wasn't entirely clear to me while reading the code. Maybe add a quick comment explaining why you introduce this.

mblondel (approved):
Thank you so much for the amazing improvements. LGTM apart from the minor comment below.

docs/notebooks/deep_learning/resnet_flax.md
fabianp: TBH I'm doing it because flax does it that way (https://github.com/google/flax/blob/main/examples/imagenet/models.py). I guess `conv: ModuleDef` is more explicit (for a human) than `conv: Any` , although the parser will treat them equally?

mblondel: If that's the Flax way, fine with me!

PR #495, opened 2023-08-07
support qr factorization in lm

amir-saadat:


tests/levenberg_marquardt_test.py
```diff
@@ -185,7 +184,7 @@ def test_double_exponential_fit_x32(self, materialize_jac, geodesic):
   # Optimization and Data Fitting" Example 6.10-. page 124. Data taken from
   # immoptibox "A MATLAB TOOLBOX FOR OPTIMIZATION AND DATA FITTING" H.B. Nielsen
   @unittest.skipIf(not jax.config.jax_enable_x64, 'test requires X64')
```
vroulet, line 186: Add jax.config.update("jax_enable_x64", True) at the start and jax.config.update("jax_enable_x64", False) at the end to be sure that the test is done.

vroulet (commented):
Thank you very much Amir! I left some comments. I'm happy to discuss more or even make some changes directly if you want.

jaxopt/_src/linear_solve.py
```diff
@@ -116,6 +123,32 @@ def solve_inv(matvec: Callable,
     raise NotImplementedError
 
 
+def solve_qr(matvec: Callable, 
+             b: jnp.ndarray, 
+             ridge: Optional[float] = None) -> tuple[jnp.ndarray, jnp.ndarray]:
```
vroulet, line 128: tuple -> Tuple (should solve failing tests)

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 """Levenberg-Marquardt algorithm in JAX."""
 
+import enum
```
vroulet, line 16: Unused

tests/levenberg_marquardt_test.py
amir-saadat: I also thought the same, and that does work for the test internal code, however, apparently, `@unittest.skipIf(...)` uses the module level environmental variables, which means that changing the setting as part of the test will not take into effect, therefore one would need to do something like the following to enforce double precision FP:

```
JAX_ENABLE_X64=True python levenberg_marquardt_test.py
```

tests/linear_solve_test.py
```diff
@@ -122,10 +124,13 @@ def matvec(x):
     x3 = linear_solve.solve_gmres(matvec, b)
     x4 = linear_solve.solve_bicgstab(matvec, b)
     x5 = linear_solve.solve_iterative_refinement(matvec, b)
+    x6 = linear_solve.solve_qr(matvec, b)
+    
     self.assertArraysAllClose(x, x2, atol=1e-4)
     self.assertArraysAllClose(x, x3, atol=1e-4)
     self.assertArraysAllClose(x, x4, atol=1e-4)
     self.assertArraysAllClose(x, x5, atol=1e-4)
+    self.assertArraysAllClose(x, x6, atol=1e-2)
```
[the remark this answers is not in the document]
amir-saadat: copy paste typo :). Fixed. I would expect to get more than this consistency for double precision, but I guess for single precision, ~1e-5 could work.

jaxopt/_src/linear_solve.py
```diff
@@ -116,6 +123,32 @@ def solve_inv(matvec: Callable,
...
+    A tuple containing Q and R matrices.
+  """
+  if ridge is not None:
+    matvec = _make_ridge_matvec(matvec, ridge=ridge)
+  if len(b.shape) == 0:
+    return b / _materialize_array(matvec, b.shape)
+  elif len(b.shape) == 1:
+    A = _materialize_array(matvec, b.shape)
+    q, r = jsp.linalg.qr(A)
+    return jsp.linalg.solve_triangular(r, q.T @ b, trans="N")
```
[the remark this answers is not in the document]
amir-saadat: you are right, 0 is the default which is the same as "N". fixed.

jaxopt/_src/linear_solve.py
```diff
@@ -116,6 +123,32 @@ def solve_inv(matvec: Callable,
...
+  else:
+    raise NotImplementedError
```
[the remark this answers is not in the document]
amir-saadat: done. (added `# TODO: support for 2D b arrays.`)

jaxopt/_src/linear_solve.py
```diff
@@ -116,6 +123,32 @@ def solve_inv(matvec: Callable,
...
+    b: RHS array.
+    ridge: optional ridge regularization.
+
+  Returns:
+    A tuple containing Q and R matrices.
+  """
+  if ridge is not None:
+    matvec = _make_ridge_matvec(matvec, ridge=ridge)
+  if len(b.shape) == 0:
+    return b / _materialize_array(matvec, b.shape)
```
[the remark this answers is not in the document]
amir-saadat: done.

jaxopt/_src/linear_solve.py
amir-saadat: done.

jaxopt/_src/levenberg_marquardt.py
amir-saadat: done.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -326,6 +329,9 @@ def update(self, params, state: NamedTuple, *args, **kwargs) -> base.OptStep:
     # Current value of the loss function F=0.5*||f||^2.
     loss_curr = state.loss
 
+    # TODO: materialize_jac currently silently use LU factorization. Needs to be 
+    # updated to use direct linear solve with passing the matrix.
+
     # For geodesic acceleration, we calculate  jtrpp=JT * r",
     # where J is jacobian and r" is second order directional derivative.
     if self.materialize_jac:
```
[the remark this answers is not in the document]
amir-saadat: there is one important difference, for the case of `materialize_jac=True`, we are materializing Jacobian, IOW, calculating the gradients using `jacfwd`. If this option is false, all variables that depend on the Jacobian like `J^T.J`, are obtained with `jvp` and `vjp` tricks, that's why you see the operators like `_jt_op`, `_jtj_op` etc. I will update the documentation so it becomes more clear.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -326,6 +329,9 @@ def update(self, params, state: NamedTuple, *args, **kwargs) -> base.OptStep:
     # Current value of the loss function F=0.5*||f||^2.
     loss_curr = state.loss
 
+    # TODO: materialize_jac currently silently use LU factorization. Needs to be
```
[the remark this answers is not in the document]
amir-saadat: please see the explanation above and I updated the docstring.

tests/linear_solve_test.py
```diff
@@ -104,8 +104,10 @@ def test_solve_dense_psd(self):
     x = linear_solve.solve_cholesky(matvec, b)
     x2 = jax.numpy.linalg.solve(A, b)
     x3 = linear_solve.solve_inv(matvec, b)
+    x4 = linear_solve.solve_qr(matvec, b)
```
vroulet, line 107: I forgot: could this be moved to solve_dense? 
QR does not require psd I think.

tests/levenberg_marquardt_test.py
vroulet: I was thinking of simply removing @unittest.skipIf(not jax.config.jax_enable_x64, 'test requires X64') in favor of adding jax.config.update("jax_enable_x64", True) at the start and jax.config.update("jax_enable_x64", False) at the end. I am not sure that the github tests for example are configured to run both in x64 and x32.

jaxopt/_src/levenberg_marquardt.py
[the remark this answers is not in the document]
vroulet: Ok, I see. The wording ``materialize_jac=True'' is confusing since even if it is false, as long as the solver is not a matrix free solver like cg or gmres (or any solver in linear_solver that does not use '_materialize_array'), the jtj will be materialized and there could be a substantial overhead (materializing the matrix can be as costly as inverting it a priori, especially for relatively low precision). I would add in the postprocess a warning if materialize_jax is false and the solver is not a matrix-free solver.
By the way the vjp is computed at several places (in _jt_op and _jtj_op). It could be shared to save some memory.

tests/levenberg_marquardt_test.py
amir-saadat: I'm also not sure how is best to do this. If we go with what you mentioned, we are basically ignoring all x32 tests (we might as well just remove them in that situation). That was actually my early on setup that I remember @mblondel suggested that we need to add x32 as it is the default usage of jaxopt. In the current setup, we are also ignoring x64 for most of the automatic integration tests, however, running x64 only needs a user specification side rather than changing the code.

tests/linear_solve_test.py
amir-saadat: :), I agree about not requiring QR, however, this is exactly what we are doing in LM. I add a test to solve_dense as well.

jaxopt/_src/levenberg_marquardt.py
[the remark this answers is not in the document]
vroulet: We may rethink the use of vjp for another CL I think. The warning is great.

tests/levenberg_marquardt_test.py
vroulet: Ok, my bad, I didn't realize the presence of the x32 test. Could you then simply make sure that `test_double_exponential_fit_x32` is also testing for different solvers?

amir-saadat: Interestingly, if we use QR solver, the algorithm doesn't converge for the x32 test. I added all the other solvers and they are fine, also fine for x64.

amir-saadat: Also, I updated the code in the materialize Jacobian to use the expected solvers and therefore removed the previous TODO and only skipping the combination that fails.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -326,53 +345,49 @@ def update(self, params, state: NamedTuple, *args, **kwargs) -> base.OptStep:
...
+      else:
+        velocity = jsp.linalg.solve(jtj_corr, state.gradient, assume_a='pos')
       delta_params = velocity
       if self.geodesic:
         rpp = (state.hess_res @ velocity) @ velocity
         # Note the same as above here that we could use inverse of jtj_corr but
         # chose to use solve for higher performance and lower numerical error.
-        acceleration = jnp.linalg.solve(jtj_corr, state.jt)
+        if self.solver == 'lu':
+          acceleration = jnp.linalg.solve(jtj_corr, state.jt)
```
vroulet, line 376: I don't see the difference between this line and line 380. 
Also, generally I would remove 'inv' since it is the least accurate and most costly algorithm. It would be useful if the inverse was kept in memory. We may keep it for now for legacy reasons.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -326,53 +345,49 @@ def update(self, params, state: NamedTuple, *args, **kwargs) -> base.OptStep:
...
       # Note that instead of taking the inverse of jtj_corr and multiply that
-      # by state.gradient, we are using `solve`, which uses LU decomposition of
+      # by state.gradient, we prefer to use `jsp.solve`, which uses Cholesky of
       # jtj_corr and uses that to obtain velocity. This has the advantage of
       # lower number of floating point operations and therefore less numerical
       # error which can be helpful for the case of single precision arithmatics.
       jtj_corr = state.jtj + damping_term
-      velocity = jnp.linalg.solve(jtj_corr, state.gradient)
+      if self.solver == 'lu':
+        velocity = jnp.linalg.solve(jtj_corr, state.gradient)
```
vroulet, line 362: Same as below: I don't get the difference between this line and line 367.

tests/linear_solve_test.py
```diff
@@ -78,8 +78,10 @@ def test_solve_dense(self):
     x = linear_solve.solve_lu(matvec, b)
     x2 = jax.numpy.linalg.solve(A, b)
     x3 = linear_solve.solve_iterative_refinement(matvec, b)
+    x4 = linear_solve.solve_qr(matvec, b)
     self.assertArraysAllClose(x, x2)
     self.assertArraysAllClose(x, x3)
+    self.assertArraysAllClose(x, x4, atol=1e-4)
```
vroulet, line 84: The fact that atol=1e-4 is needed is strange (I agree that it indeed does not converge without it. We should report a bug to the JAX team.

tests/levenberg_marquardt_test.py
vroulet: Ok, we may want to end this PR, so I am ok to add a todo and a warning and add an enhancement issue to review this later. We should also start having benchmarks.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -326,70 +368,73 @@ def update(self, params, state: NamedTuple, *args, **kwargs) -> base.OptStep:
...
+      if self.solver == 'qr':
+        damping_term = jnp.sqrt(state.damping_factor) * jnp.identity(params.size)
+        aug_jac = jnp.vstack([state.jac, damping_term])
+        aug_res = jnp.hstack([state.residual, jnp.zeros(params.size)])
+        q, r = jsp.linalg.qr(aug_jac, mode='economic')
+        velocity = jsp.linalg.solve_triangular(r, q.T @ aug_res)
+      else:
+        damping_term = state.damping_factor * jnp.identity(params.size)
+        jtj_corr = state.jtj + damping_term
+        if self.solver == 'lu':
```
vroulet, line 391: Thanks for all the work Amir! 
I'm really sorry but I'm still confused about the difference between the 'lu' and the 'inv' case in this if statement. The lines 392: `velocity = jnp.linalg.solve(jtj_corr, state.gradient)` and 394 `velocity = jnp.linalg.solve(jtj_corr, state.gradient)` are identical to me. If the difference lies in the way state.gradient or jtj_corr are defined, it would be better to add a note about it to ease the reading.

vroulet:
Thank you very much Amir for the hard work! The resolution of the subproblem by using directly the rectangular matrix is very nice. I left you a (hopefully last) comment as I didn't understand your previous answer and I wanted to point to the new code.

jaxopt/_src/levenberg_marquardt.py
vroulet: Great! Thank you, I would not be surprised if the inv does not give very accurate solutions.

vroulet:
Great! I think we are good to go! Could you squash the commits to finalize the PR?
Thank you again for all the work!

mblondel:
The issue is that JAX uses float32 by default and the rest of JAXopt works in float32. So ideally, LM with default options should work in float32 precision.

vroulet:
The error can be reproduced locally using as initialization: `onp.asarray([40, 50, 30], dtype=float)`.
Then investigating the results step by step using `verbose=True` reveals that the damping factor blows up near the minimum:
```
   Iteration        Cost       ||Gradient||  Damping Factor 
       1         5.9925e+06     7.4510e+06      1.452e+07   
       2         3.6550e+06     4.4647e+06      8.263e+06   
       3         2.1990e+06     2.6407e+06      4.776e+06   
       4         1.3196e+06     1.5599e+06      2.76e+06    
       5         7.8952e+05     9.2045e+05      1.593e+06   
       6         4.7060e+05     5.4237e+05      9.18e+05    
       7         2.7913e+05     3.1899e+05      5.28e+05    
       8         1.6451e+05     1.8713e+05      3.029e+05   
       9         9.6168e+04     1.0942e+05      1.732e+05   
      ...
      80         4.3963e-05     1.5993e-02       0.05916    
      81         4.3963e-05     1.5993e-02        1.893     
      82         4.3910e-05     5.1219e-03        0.631     
      83         4.3910e-05     5.1219e-03        1.262     
      84         4.3910e-05     5.1219e-03        5.048     
      85         4.3910e-05     5.1219e-03        40.39     
      86         4.3910e-05     5.1219e-03        646.2     
      87         4.3910e-05     5.1219e-03      2.068e+04   
      88         4.3910e-05     5.1219e-03      1.323e+06   
      89         4.3910e-05     5.1219e-03      1.694e+08   
      90         4.3910e-05     5.1219e-03      4.336e+10   
      91         4.3910e-05     5.1219e-03      2.22e+13    
      92         4.3910e-05     5.1219e-03      2.274e+16   
      93         4.3910e-05     5.1219e-03      4.656e+19   
      94         4.3910e-05     5.1219e-03      1.907e+23   
      95         4.3910e-05     5.1219e-03      1.562e+27   
      96         4.3910e-05     5.1219e-03      2.56e+31    
      97         4.3910e-05     5.1219e-03      8.388e+35   
      98         4.3910e-05     5.1219e-03         inf      
      99             nan            nan            inf      
```
This seems to mean that there is a bug in the algorithm for this to happen. One may for example add a maximal value for the damping factor. Note that stopping the algorithm at 98 iterations enable the test to pass. 

PS: Trying to add `  jax.config.update('jax_enable_x64', True)` before `absltest.main()` still skips the tests in x64 on my side. On the other hand simply switching the precision at the start of the test seems to ensure that all tests pass, namely, using 
```
  @parameterized.product(materialize_jac=[True, False])
  def test_rosenbrock_x64_materialize_jac(self, materialize_jac):
    jax.config.update('jax_enable_x64', True)
    lm = LevenbergMarquardt(
        residual_fun=_rosenbrock,
        damping_parameter=1e-3,
        stop_criterion='madsen-nielsen',
        xtol=1e-12,
        gtol=1e-8,
        maxiter=16,
        materialize_jac=materialize_jac)
    optimize_info = lm.run(onp.array([-1.2, 1]))

    self.assertArraysAllClose(
        optimize_info.params - onp.array([1, 1]),
        onp.array([-4.07e-09, -8.16e-09]),
        atol=1e-11)
    self.assertAllClose(optimize_info.state.iter_num, 16)
    jax.config.update('jax_enable_x64', False)
 ```

vroulet:
I added a maximal value for the increase factor and the damping to avoid the aforementioned behavior and avoid NaNs.
I also added a minimal value for the denominator of the fraction computing the gain (I also used the absolute value of the denominator to be sure). 
I made sure that all tests are both in x32 and x64 and added commented lines in the test file to be able to test easily in X64 (the double precision needs to be enabled before the definition of the class).
I readded the test of double exponential fit in X32. It fails on github. For now, I commented it and I would probably like to review the full code in another PR if that's ok with you @amir-saadat and @mblondel. My idea is to have a generalized gauss-newton class that encompasses GN, Levenberg-Marquardt and some robust variants (see https://github.com/google/jaxopt/discussions/484).
In any case, thank you very much for all the work Amir!

vroulet:
> I think this is not necessary, and in fact, I don't know if it is correct to take the `abs` of the denom, the definition is something like this `gain_ratio ~ dLoss / dLoss_approx` and we can be changing the sign of the denom with that. I reverted that change, and it seems like the tests are still passing. Please let me know if you still think that we should use it if I missed something.

Theoretically I think that dLoss_approx is necessarily non-negative. Clearly you don't want dLoss_approx to be potentially negative, hence the absolute value just in case. Otherwise you end up accepting points for which dLoss < 0 where the next step is worse than the current and it should not be accepted. In any case, if dLoss and dLoss_approx tend to 0 we reach a situation of the form 0/0 which a computer does not necessarily handle well (potential NaNs). 

> not sure if i'm seeing that part on my end. I only see this unused line `self.test_in_x_64 = True`. I think the skip test is a module level env variable. so you would have to set that up at the python call time.

See line 73 to 75 in `levenberg_marquardt_test.py`. I think `self.test_in_x_64 = True` refers to a previous commit that was changed. I added this to facilitate the test in x64 locally. You may have had another solution, but in any case, documenting it is a good idea. Anyway, the code needs to work in x32 and it appears that further work (in a separate CL) would be needed for that. 

The tests pass for now so I close this PR and I'll open a new one to have a closer look at the code later.
Again thanks for your work Amir!

PR #498, opened 2023-08-14
fix has_aux bug in gauss_newton

amir-saadat:
@mblondel @vroulet , 
Making some improvements and bug fixes for GaussNewton. In particular, I'm making some minor changes in the printing error section so that we can Jit functions while using verbose=True. This should also let other solvers be jittable while using verbose=True.

### DIFF
--- a/jaxopt/_src/gauss_newton.py
+++ b/jaxopt/_src/gauss_newton.py
@@ -33,6 +33,7 @@ class GaussNewtonState(NamedTuple):
   """Named tuple containing state information."""
   iter_num: int
   residual: Any
+  value: Any
   delta: Any
   error: float
   gradient: Any
@@ -55,6 +56,7 @@ class GaussNewton(base.IterativeSolver):
     implicit_diff: whether to enable implicit diff or autodiff of unrolled
       iterations.
     implicit_diff_solve: the linear system solver to use.
+    has_aux: whether ``residual_fun`` outputs auxiliary data or not.
     verbose: whether to print error on every iteration or not.
     jit: whether to JIT-compile the bisection loop (default: "auto").
     unroll: whether to unroll the bisection loop (default: "auto").
@@ -62,10 +64,10 @@ class GaussNewton(base.IterativeSolver):
   residual_fun: Callable
   maxiter: int = 30
   tol: float = 1e-5
-  verbose: bool = False
   implicit_diff: bool = True
   implicit_diff_solve: Optional[Callable] = None
   has_aux: bool = False
+  verbose: bool = False
   jit: base.AutoOrBoolean = "auto"
   unroll: base.AutoOrBoolean = "auto"
 
@@ -90,6 +92,7 @@ class GaussNewton(base.IterativeSolver):
     return GaussNewtonState(iter_num=jnp.asarray(0),
                             error=jnp.asarray(jnp.inf),
                             residual=residual,
+                            value=0.5 * jnp.sum(jnp.square(residual)),
                             delta=init_params,
                             gradient=gradient,
                             aux=aux)
@@ -108,7 +111,6 @@ class GaussNewton(base.IterativeSolver):
       (params, state)
     """
     residual, aux = self._fun_with_aux(params, *args, **kwargs)
-
     matvec = lambda v: self._jtj_op(params, v, *args, **kwargs)
     gradient = self._jt_op(params, residual, *args, **kwargs)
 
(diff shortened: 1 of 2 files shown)


vroulet:
Hello Amir, 
Sorry for the delay I wanted the previous PR on LM to be done. 
Concerning this one:
- Using jax.debug for the prints is great. It would probably be best to have this done outside of this PR and review all algorithms. For example, one may simplify the printing of the errors in the zoom linesearch too and revise all algorithms that give additional information when verbose=True. I can do that if you don't have the bandwidth.
- For Gauss-Newton, I wasn't there when it was first implemented and I would have some generic questions before enhancing e.g. the jit possibilities:
  - Gauss-Newton *needs* a linesearch to converge otherwise it is well-known that it can diverge (see Wikipedia counter-example for example). 
  - I don't really see any use-case for a GN method if we have a LM method that is functional. True Gauss-Newton has a quadratic local convergence but globally it can be pretty bad. On the other hand, LM can better behave globally. True LM with fixed regularization parameter only converges linearly but with varying regularization, I believe the quadratic convergence can be ensured. Overall I think LM is more general anyway. What do you think?

vroulet:
> Sounds good, I moved it to https://github.com/google/jaxopt/pull/503

Thanks, that'll be a very useful addition!

> This would then become an LM variant right?

Not exactly, LM uses a trust-region method. I was saying a simple linesearch. By the way, it would be interesting to have a generic trust-region method which could apply not only to LM.

> I agree with you that LM would be preferred than GN ...

Ok! So shall we close this PR then for now if that's ok with you?

jaxopt/_src/gauss_newton.py
```diff
@@ -119,6 +120,7 @@ def update(self,
     state = GaussNewtonState(iter_num=state.iter_num + 1,
                              error=tree_l2_norm(delta_params),
                              residual=residual,
+                             loss=0.5 * (residual @ residual),
```
vroulet, line 123: loss -> value (that's the current name given in other solvers).

jaxopt/_src/gauss_newton.py
```diff
@@ -91,6 +92,7 @@ def init_state(self,
     return GaussNewtonState(iter_num=jnp.asarray(0),
                             error=jnp.asarray(jnp.inf),
                             residual=residual,
+                            loss=0.5 * (residual @ residual),
```
vroulet, line 95: loss -> value

jaxopt/_src/gauss_newton.py
```diff
@@ -33,6 +33,7 @@ class GaussNewtonState(NamedTuple):
   """Named tuple containing state information."""
   iter_num: int
   residual: Any
+  loss: Any
```
vroulet, line 36: loss -> value

jaxopt/_src/gauss_newton.py
vroulet, line 123: As you're on it, make it able to handle scalars as you did for LM.

tests/gauss_newton_test.py
```diff
@@ -69,6 +69,21 @@ def setUp(self):
     ])
     self.init_temperature_record_coeffs = onp.array([10, 0.5, 10.5, 50])
 
+  def test_aux_true(self):
```
vroulet, line 72: Make it a one step verification (like max_iter =2) with a fixed aux for any state (like True) and check if the aux is still the same. I think other solverrs have such tests. The current tests are already taking a long time so it's preferable to minimize the lengths of the tests if we can.

tests/gauss_newton_test.py
```diff
@@ -80,8 +95,8 @@ def test_enzyme_reaction_parameter_fit(self):
         self.rate_data)
 
     self.assertArraysAllClose(optimize_info.params,
-                              onp.array([0.36183689, 0.55626653]))
-
+                              onp.array([0.36183689, 0.55626653]),
```
vroulet, line 98: Where does this value come from? 
- if you have a ref add it
- otherwise, 
  - you may simply compare gn to a solver like LBFGS
  - you may create artifically some nonlinear data, add some noise then compare GN to the solution of LBFGS initialized at the parameter that generated the data.
  
Note that this is optional, that is, since we may redo this class later, we can pass that. But if you have a ref this will be helpful later.

tests/gauss_newton_test.py
```diff
@@ -115,8 +130,8 @@ def test_temperature_record_four_parameter_fit(self):
     # Checking against the expected values
     self.assertArraysAllClose(
         optimize_info.params,
-        onp.array([16.63994555, 0.46327812, 10.85228919, 76.19086103]))
-    self.assertAllClose(optimize_info.state.iter_num, 30)
+        onp.array([16.63994555, 0.46327812, 10.85228919, 76.19086103]),
```
vroulet, line 133: Same as above, a ref would be good if you have one.

vroulet:
Thanks for the PR! I left you some comments.

vroulet:
Hello Amir, 
If the comments above were fixed can you mark them as resolved and squash everything? 
I'll add the pull ready tag.
Thanks for all!

PR #502, opened 2023-08-16
fix_optax_wrapper

vroulet:
The optax wrapper was not handling value_and_grad properly anymore due to the check of dtype in value (precisely at `value=jnp.asarray(jnp.inf, value.dtype)` where value was a tuple when `value_and_grad=True`) .
Here is a MWE. The fix is simple, I can add a test and review all similar calls of dtype of value if you want. (I needed this fix for a personal project but I can workaround).
```
import jax
from jax import numpy as jnp
from jaxopt import OptaxSolver
import optax

def fun(x): 
  return 0.5*x**2
val_and_grad_fun = jax.value_and_grad(fun)
solver = OptaxSolver(val_and_grad_fun, opt=optax.sgd(learning_rate=0.1), value_and_grad=True)
solver.run(jnp.asarray(1.))
# AttributeError: 'tuple' object has no attribute 'dtype'
```

### DIFF
--- a/jaxopt/_src/armijo_sgd.py
+++ b/jaxopt/_src/armijo_sgd.py
@@ -248,11 +248,7 @@ class ArmijoSGD(base.StochasticSolver):
     else:
       velocity = tree_zeros_like(init_params)
 
-    if self.has_aux:
-      value, aux = self.fun(init_params, *args, **kwargs)
-    else:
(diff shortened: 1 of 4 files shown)


mblondel:
Thanks. A non-regression test would indeed be great to ensure that we don't reintroduce the bug silently in the future.

mblondel (approved):
LGTM

PR #504, opened 2023-08-18
support scalar output in lm

amir-saadat:
this is just a simple update to the code to ensure that scalar function output are supported. Although less of interest, might be considered an edge case.

### DIFF
--- a/jaxopt/_src/levenberg_marquardt.py
+++ b/jaxopt/_src/levenberg_marquardt.py
@@ -43,7 +43,7 @@ class LevenbergMarquardtState(NamedTuple):
   damping_factor: float
   increase_factor: float
   residual: Any
-  loss: Any
+  value: Any
   delta: Any
   error: float
   gradient: Any
@@ -223,7 +223,7 @@ class LevenbergMarquardt(base.IterativeSolver):
         increase_factor=2,
         error=tree_l2_norm(gradient),
         residual=residual,
-        loss=0.5 * jnp.sum(jnp.square(residual)),
+        value=0.5 * jnp.sum(jnp.square(residual)),
         delta=delta_params,
         gradient=gradient,
         jac=jac,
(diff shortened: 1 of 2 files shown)


vroulet:
Hello Amir!
Thanks for that! Did you check that the current code did not pass your test?

vroulet:
Ok I checked. Indeed the current code does not pass the test.
In the future, it may be better not to use operators such as `@` and keep jax/numpy operators, or, even better, make sure that the code can work with pytrees.

jaxopt/_src/levenberg_marquardt.py
```diff
@@ -228,7 +228,7 @@ def init_state(self, init_params: Any, *args,
         increase_factor=2,
         error=tree_l2_norm(gradient),
         residual=residual,
-        loss=0.5 * (residual @ residual),
+        loss=0.5 * jnp.sum(jnp.square(residual)),
```
vroulet, line 231: Since you are at it, loss -> val. Same at other places in the code.

tests/levenberg_marquardt_test.py
```diff
@@ -316,6 +315,14 @@ def test_has_aux_true(self):
     self.assertAllClose(x_opt, x_gt, atol=1e-6)
     self.assertAllClose(state.aux, x_gt**2, atol=1e-6)
 
+  def test_scalar_output_fun(self):
+    lm = LevenbergMarquardt(
+        residual_fun=lambda x: x @ x,
+        tol=1e-6,)
```
vroulet, line 321: tol=1e-1

tests/levenberg_marquardt_test.py
```diff
@@ -316,6 +315,14 @@ def test_has_aux_true(self):
...
+    x_init = jnp.ones((10,))
```
vroulet, line 322: jnp.ones((2,))

tests/levenberg_marquardt_test.py
```diff
@@ -316,6 +315,14 @@ def test_has_aux_true(self):
...
+    x_opt, _ = lm.run(x_init)
+
+    self.assertAllClose(x_opt, jnp.zeros((10,)), atol=1e-2)
```
vroulet, line 325: jnp.zeros((2,)) 
atol=1e0

vroulet (approved):

PR #507, opened 2023-08-23
Add non-jittable fori_loop

### LINKED ISSUE #462: BlockCD does not support non-jitted functions
For now because `fori_loop` automatically compiles the function, BlockCD does not support non-jitted functions.
The resolution should be very similar to what was done in #449 .

ayush-1506:
The nomenclature might not be ideal since it's easy to confuse `from jaxopt._src.fori_loop.fori_loop` for `jax.lax.fori_loop`. Let me know if that can be a problem and I'll rename it. I've also not added any tests at the moment - can add those after an initial review.

Hopefully fixes #462

### DIFF
--- a/jaxopt/_src/block_cd.py
+++ b/jaxopt/_src/block_cd.py
@@ -28,6 +28,7 @@ import jax
 import jax.numpy as jnp
 
 from jaxopt._src import base
+from jaxopt._src.fori_loop import fori_loop
 from jaxopt._src import implicit_diff as idf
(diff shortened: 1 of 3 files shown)


mblondel:
Thanks for the PR and sorry for the late reply! Since we already have `loop.py`, I would move your function there, instead of creating a new file. Also we'll need two tests: one in `loop_test.py` for the function that you introduce and one in `block_cd_test.py` for checking that #462 is indeed fixed.

PR #503, opened 2023-08-17
jittable verbose=True

amir-saadat:
updated the print calls of the optimizers to `jax.debug.print(..)` so it can be jittable for closure definitions.

PR #512, opened 2023-08-31
Jit update method and remove jit="auto".

mblondel:
This pull-request adds support for automatically jit compiling the `update` method. It also removes `jit="auto"`, which was made unnecessary thanks to the jittable verbose support, added in #503.


fabianp (approved):
Looks good to me!

PR #513, opened 2023-08-31
Capture stdout in verbose test.

mblondel:


vroulet (approved):
Thank you! I didn't know we could avoid the prints to be displayed during tests.

PR #526, opened 2023-09-04
Build doc as part of the CI workflow

fabianp:


mblondel:
Thanks a lot for doing this! What kind of failures does this detect currently? Is it mainly syntax errors in the examples and notebooks? I guess it doesn't run examples in the narrative documentation and in the docstrings yet?

fabianp:
it currently runs things under example/ and notebooks/ (unless those notebooks that have been excluded bc they take too long). This allowed me to catch the 3 broken examples in https://github.com/google/jaxopt/pull/525 , and the missing requirements added to this PR.

Next steps would be IMO (to be done in a different PR):

  *  to run doctests, which I would do with pytest, not sphinx, for example using https://docs.pytest.org/en/7.3.x/how-to/doctest.html
  * use this build of the narrative documentation to push to https://jaxopt.github.io/dev/ , so we can see the doc for the development version

mblondel:
Thanks. It just wanted to test understand exactly what it tests. Completely agreed that the rest should be done separately.

mblondel (approved):

PR #529, opened 2023-09-10
added the computed initialization of the jac approx in Broyden

zaccharieramzi:
WIP: need to run the existing tests and add the previously failing test written by @vroulet in a [previous PR](https://github.com/google/jaxopt/pull/521#issuecomment-1703639170)

In order to implement this change, I had to change the dtype of `gamma` which is now a pytre rather than a scalar, since it can be different for each leaf of the parameters.


zaccharieramzi:
So actually the failure that @vroulet was showcasing [here](https://github.com/google/jaxopt/pull/521#issuecomment-1703639170) is actually already a [test](https://github.com/google/jaxopt/blob/main/tests/broyden_test.py#L56) in the broyden test suite.

Everything is green now (with simpler settings), so this is ready for review.

PR #535, opened 2023-09-16
added jax tests and tested in x64, increased maxls for bfgs, lbfgs

vroulet:
Adding tests from jax to enable moving jax.optimize to using jaxopt, see https://github.com/google/jax/pull/17025. 
Tests failed with previous setup of maximal number of linesearch iterations fixed to 15 and pass by fixing it to 30.
Also, checked if lbfgs and bfgs tests could pass in x64 and added modifications and notes where necessary.

Note: an alternate fix would be to let the zoom linesearch make a step even it has not found a sufficient decrease. The failures were due to the fact that contrarily to scipy we do not "hope for the best" if the linesearch fails. But then this makes the optimizer possibly jump outside the domain of the objective (nan or inf). As long as we print a warning in this case (even without verbose), this could make sense too. Let me know what you prefer.

### DIFF
--- a/jaxopt/_src/zoom_linesearch.py
+++ b/jaxopt/_src/zoom_linesearch.py
@@ -37,13 +37,25 @@ from jaxopt.tree_util import tree_conj
 # pylint: disable=g-bare-generic
 # pylint: disable=invalid-name
 
-# Flags are encoded in base 2 to enable bitwise logic
-_FLAG_NOT_DESCENT_DIRECTION = 1
-_FLAG_CURVATURE_COND_NOT_SATSIFIED = 2
-_FLAG_MAX_ITER_REACHED = 4
-_FLAG_INTERVAL_TOO_SMALL = 8
-_FLAG_NAN_INF_VALUES = 16
-_FLAG_NO_STEPSIZE_FOUND = 32
+# Flags to print errors, used in tests
+WARNING_PREAMBLE = \
+  "ZoomLineSearchWarning: "
+FLAG_NOT_A_DESCENT_DIRECTION = WARNING_PREAMBLE + \
+  "Provided linesearch direction is not a descent direction. " + \
+  "The linesearch will probably fail."
+FLAG_NAN_INF_VALUES = WARNING_PREAMBLE + \
+  "NaN or Inf values encountered in function values."
+FLAG_INTERVAL_NOT_FOUND = WARNING_PREAMBLE + \
+  "No interval satisfying curvature condition. " + \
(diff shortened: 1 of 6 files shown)


mblondel:
Thank you @vroulet! So in the end, do you prefer to increase the number of maximum iterations from 15 to 30, or, do you prefer to force to take a step and "hope for the best"? CC @fabianp

fabianp:
increasing the max ls iterations seems safer to me, but no strong preference either way

vroulet:
I've quickly tried the option of just letting it go (taking last stepsize computed). It does not solve for example the issue of scipy and jaxopt having different behaviors on the himmelblau problem with bad initialization (again both jaxopt and scipy find a global minima, they are simply different so from an optimization viewpoint it's not an issue).

I looked a bit more into the default parameters of scipy. The "hope for the best" option was when using a linesearch associated to a nonlin problem not the linesearch associated to unconstrained optimizers like bfgs. For the later, the default option for the zoom linesearch is 100 (see https://github.com/scipy/scipy/blob/d5a1db67a041ffeb2358c0d563dbe2e123815e9d/scipy/optimize/_linesearch.py#L332). So I think it's safe to increase the number of linesearch iterations. We may investigate why scipy and jaxopt differ on the himmelblau problem with bad initialization and/or add an option for letting it go later. 

Finally I made a commit for the zoom linesearch to always print a warning if no stepsize has been found. And I simplified largely the printing of warnings (no bitewise coding, just jax.debug.print), this makes the code more readable (bitwise logic differs between jax and numpy so it could be confusing at first glance.

If that looks good to you I'll squash.

jaxopt/_src/zoom_linesearch.py
```diff
@@ -37,13 +37,25 @@
...
 
-# Flags are encoded in base 2 to enable bitwise logic
-_FLAG_NOT_DESCENT_DIRECTION = 1
-_FLAG_CURVATURE_COND_NOT_SATSIFIED = 2
-_FLAG_MAX_ITER_REACHED = 4
-_FLAG_INTERVAL_TOO_SMALL = 8
-_FLAG_NAN_INF_VALUES = 16
-_FLAG_NO_STEPSIZE_FOUND = 32
+# Flags to print errors, used in tests
+_WARNING_PREAMBLE = \
```
mblondel, line 41: Minor detail: since we are in `_src/` it's ok if you don't prefix all the constants with a leading `_`.

mblondel (approved):
LGTM apart from minor remark below.

PR #541, opened 2023-09-23
Enhanced zoom linesearch with earlier stopping, actionable messages, …

vroulet:
I discussed with Brendan McMahan and Ryan McKenna, who are using JAXOpt and especially LBFGS. They pointed out some potential improvements. The main one being the actionable messages. I inferred an earlier stopping in case the interval is too small from the colab they shared.

So this PR is about enhancing zoom with
- Earlier stopping: if the interval in which zoom searches is reduced by below some threshold and a sufficient decrease has been found for some stepsize s, just take s. This avoids wasting time finding a curvature condition while a good stepsize ensuring a sufficient decrease has been found. That threshold is now an option. If it appears useful to lower it, I'll document it.
- Various actionable messages if the lineseach fails.
- Decided to let the linesearch use stepsize even if not giving sufficient decrease with a clear warning, except if that stepsize would lead to NaNs or Infs and it that case the algorithm stays stuck (with a clear message).

I've also done the following (can move to another PR if you want).
- Decoupled max_stepsize of initial guess for linesearches to max_stepsize actually taken (documented it). Zoom may use larger stepsize than 1 to satisfy the curvature, however, 1 is a good maximal guess for the stepsize. Such a strategy appeared efficient on the hard zakharov problem.
- For HagerZhangLinesearch, max_stepsize is actually not used in the algorithm. For Backtracking, a priori the algorithm itself should never increase the stepsize beyond its guess, so I thought it would be better to let max_stepsize be only handled in the initial guess.

### DIFF
--- a/jaxopt/_src/bfgs.py
+++ b/jaxopt/_src/bfgs.py
@@ -101,7 +101,7 @@ class BFGS(base.IterativeSolver):
     increase_factor: factor by which to increase the stepsize during line search
       (default: 1.5).
     max_stepsize: upper bound on stepsize.
-    min_stepsize: lower bound on stepsize.
+    min_stepsize: lower bound on stepsize guess at start of the linesearch run.
     implicit_diff: whether to enable implicit diff or autodiff of unrolled
       iterations.
     implicit_diff_solve: the linear system solver to use.
--- a/jaxopt/_src/linesearch_util.py
+++ b/jaxopt/_src/linesearch_util.py
@@ -58,12 +58,12 @@ def _setup_linesearch(
         verbose=verbose,
     )
   elif linesearch == "hager-zhang":
+    # NOTE(vroulet): max_stepsize has no effect in HZ
     linesearch_solver = HagerZhangLineSearch(
         fun=fun,
         value_and_grad=value_and_grad,
         has_aux=has_aux,
         maxiter=maxlsiter,
-        max_stepsize=max_stepsize,
         jit=jit,
         unroll=unroll,
         verbose=verbose,
--- a/jaxopt/_src/broyden.py
+++ b/jaxopt/_src/broyden.py
@@ -165,7 +165,7 @@ class Broyden(base.IterativeSolver):
     increase_factor: factor by which to increase the stepsize during line search
       (default: 1.5).
     max_stepsize: upper bound on stepsize.
-    min_stepsize: lower bound on stepsize.
+    min_stepsize: lower bound on stepsize guess at start of the linesearch run.
 
     history_size: size of the memory to use.
     gamma: the initialization of the inverse Jacobian is going to be gamma * I.
--- a/jaxopt/_src/hager_zhang_linesearch.py
+++ b/jaxopt/_src/hager_zhang_linesearch.py
@@ -82,7 +82,7 @@ class HagerZhangLineSearch(base.IterativeLineSearch):
     c1: constant used by the Wolfe and Approximate Wolfe condition.
     c2: constant strictly less than 1 used by the Wolfe and Approximate Wolfe
       condition.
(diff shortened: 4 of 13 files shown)


jaxopt/_src/bfgs.py
```diff
@@ -100,8 +100,14 @@ class BFGS(base.IterativeSolver):
       backtracking line search (default: 0.8).
     increase_factor: factor by which to increase the stepsize during line search
       (default: 1.5).
-    max_stepsize: upper bound on stepsize.
-    min_stepsize: lower bound on stepsize.
+    max_stepsize: upper bound on stepsize guess at start of each linesearch run
+      for linesearch_init='increase'.
+      Note that the linesearch is allowed to take a larger stepsize to satisfy
```
mblondel, line 105: It's a bit strange to say that. Maybe we need to make the distinction between `init_stepsize` and `max_stepsize` then? Also this remark is only for zoom ?

jaxopt/_src/bfgs.py
```diff
@@ -286,7 +292,7 @@ def __post_init__(self):
         value_and_grad=True,
         has_aux=True,
         maxlsiter=self.maxls,
-        max_stepsize=self.max_stepsize,
+        max_stepsize=None,
```
mblondel, line 295: Why did you change this line?

jaxopt/_src/bfgs.py
vroulet: - Yes, I can add a max_init_stepsize and keep the max_stepsize as it is. Max_stepsize would really enforce a maximum stepsize as expected. The min_stepsize does not enforce a minimum stepsize either during the search. So we may do the same behavior (add a min_init_stepsize and let a min_stepsize intervenes in the linesearches, this would let the users have the possibility to enforce the linesearches to jump if the minimum stepsize is reached and no stepsize has been found).
- Yes, only zoom may increase the initial stepsize during its search (Backtracking always reduces and HagerZhang ignores max_stepsize parameter).

jaxopt/_src/bfgs.py
vroulet: I thought the max_stepsize was not needed a priori but I can differentiate max_init_stepsize and max_stepsize as said in the other document. If that sounds good to you I'll put back max_stepsize=self.max_stepsize here then.

mblondel: It is confusing for me if we have `max_stepsize` option but we do not propagate it...

jaxopt/_src/linesearch_util.py
```diff
@@ -33,15 +33,16 @@ def _setup_linesearch(
...
+  # and HagerZhang. Max_stepsize is handled at stepsize guess 
+  # in init_stepsize
   available_linesearches = ["backtracking", "zoom", "hager-zhang"]
   if linesearch == "backtracking":
     linesearch_solver = BacktrackingLineSearch(
         fun=fun,
         value_and_grad=value_and_grad,
         has_aux=has_aux,
         maxiter=maxlsiter,
-        max_stepsize=max_stepsize,
```
mblondel, line 44: Put back?

jaxopt/_src/bfgs.py
vroulet: I kept the solvers as they were. So kept max_stepsize. I've simply changed the min_stepsize doc.

jaxopt/_src/bfgs.py
vroulet: Kept as it was.

jaxopt/_src/linesearch_util.py
vroulet: Done

mblondel (approved):

Issue #531, opened 2023-09-11
Number of gradient evaluations not applicable to scipy's gradient-free optimizers


dominikstrb:
Since commit [#28a474e](https://github.com/google/jaxopt/commit/28a474e307665971ce3e535d1984325e0ec5b1e4), scipy optimizers that do not use gradients do not work from `jaxopt.ScipyMinimize` anymore.

Because `res.njev` is accessed on the result regardless of the `method` argument, you get an 

```
AttributeError: njev. Did you mean: 'nfev'?
```

if you run `jaxopt.ScipyMinimize` e.g. with Nelder-Mead or Powell.

PR #542, opened 2023-10-03
made sure to use njev only when available in scipy optimize results

### LINKED ISSUE #536: ScipyMinimize(method="TNC") is broken
Scipy's TNC implementation does not set the number of Jacobian evaluations  in the `ScipyOptimizeResult`, which causes the jaxopt wrapper to throw:
```
ScipyMinimize(method="TNC", fun=lambda x: x**2).run(1.0)
# Traceback (most recent call last):     
# ...
# KeyError: 'njev'
```
I'm running jaxopt 0.8 and Scipy 1.11.2.

zaccharieramzi:
This should fix #536 and #531 

Basically makes the `nfev` (function eval), `njev` (jacobian eval) and `nhev` (hessian eval) counts 0 if they are not present in the scipy results.

This was already the behaviour for the hessian.

### DIFF
--- a/jaxopt/_src/scipy_wrappers.py
+++ b/jaxopt/_src/scipy_wrappers.py
@@ -359,19 +359,21 @@ class ScipyMinimize(ScipyWrapper):
     else:
       hess_inv = None
 
-    try:
-      num_hess_eval = jnp.asarray(res.nhev, base.NUM_EVAL_DTYPE)
-    except AttributeError:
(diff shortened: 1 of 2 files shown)


mblondel:
Thanks a lot for the fix @zaccharieramzi. Can you add a non-regression test, for example using "TNC" as in #536.

zaccharieramzi:
Sure! Will do

zaccharieramzi:
@mblondel : added nonregression tests, made sure that they fail without the fix and they are green now

PR #544, opened 2023-10-09
expanded verbose for all algorithms, made verbose a boolean for all algorithms

vroulet:
Replaced "Error" by specific name of error considered and added other information when verbose=True.
Should be useful for debugging more quickly.
All verbose are now boolean too (some were int).

### DIFF
--- a/jaxopt/_src/armijo_sgd.py
+++ b/jaxopt/_src/armijo_sgd.py
@@ -21,6 +21,7 @@ from typing import Any
 from typing import Callable
 from typing import NamedTuple
 from typing import Optional
+from typing import Union
 
 import jax
 import jax.lax as lax
@@ -191,7 +192,7 @@ class ArmijoSGD(base.StochasticSolver):
     maxiter: maximum number of solver iterations.
     maxls: maximum number of steps in line search.
     tol: tolerance to use.
-    verbose: whether to print error on every iteration or not.
+    verbose: whether to print information on every iteration or not.
 
     implicit_diff: whether to enable implicit diff or autodiff of unrolled
       iterations.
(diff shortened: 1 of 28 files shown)


jaxopt/_src/armijo_sgd.py
```diff
@@ -224,7 +224,7 @@ class ArmijoSGD(base.StochasticSolver):
   maxiter: int = 500
   maxls: int = 15
   tol: float = 1e-3
-  verbose: int = 0
+  verbose: bool = False
```
mblondel, line 227: The idea of using an `int` was that we could have several levels of verbosity:
- 1: verbose
- 2: very verbose
- 3: very very verbose

For example, 2 could be used to print linesearch information.

Do you think it's unnecessary?

jaxopt/_src/armijo_sgd.py
```diff
@@ -315,6 +315,15 @@ def update(self, params, state, *args, **kwargs) -> base.OptStep:
                              stepsize=jnp.asarray(stepsize, dtype=dtype),
                              velocity=next_velocity)
 
+    if self.verbose:
+      self.log_info(
+          next_state,
+          error_name="Grad. Norm",
+          additional_info={
+              'Obj. Val.': next_state.value,
```
mblondel, line 323: I would avoid abbreviations, e.g., I prefer `Objective value` here.

vroulet: Done

jaxopt/_src/armijo_sgd.py
vroulet: I did a two level verbose for the solvers with linesearch as you suggested. 
I made all verbose types to be boolean or int to be consistent.
Thanks for the suggestion!

mblondel (approved):
Thanks Vincent

PR #556, opened 2023-11-22
fix deprecated load_boston from scikitlearn

vroulet:
`load_boston` has been removed from scikit-learn since version 1.2 for ethical reasons.
This should fix the issue for the docs to build.


fabianp (approved):
thanks!

PR #559, opened 2023-12-03
Linesearch in Broyden

zaccharieramzi:
This PR follows an email thread with @vroulet (sorry for taking this long to address this).
Basically the idea is that at the moment the Broyden tests are passing but with failed line searches.
Ideally, they should pass without the linesearch failing.

This PR makes sure that at least one test is testing that condition, in addition to using the more modern linesearch API for Broyden.

At the moment, the test is failing because we cannot modify the `condition` parameter of the linesearch. Indeed, as pointed out by @vroulet , scipy uses the armijo line search for Broyden, and if we hack our way into using it, the test passes. 
@fabianp I noticed you deprecated `condition` [in this commit](https://github.com/google/jaxopt/commit/7238f760c717e1db534fae3890150b0bfed82a7d), what was the reason for this?

This needs to be merged after #529

### DIFF
--- a/tests/broyden_test.py
+++ b/tests/broyden_test.py
@@ -48,7 +48,7 @@ class BroydenTest(test_util.JaxoptTestCase):
       return jnp.sin(x[0]) * (x[0] ** 2) - x[0], x[1] ** 3 - x[1]
     x0 = jnp.array([0.6, 0., -0.1]), jnp.array([[0.7], [0.5]])
     tol = 1e-6
-    sol, state = Broyden(g, maxiter=1000, tol=tol, jit=jit, stop_if_linesearch_fails=True, linesearch="backtracking", maxls=30).run(x0)
+    sol, state = Broyden(g, maxiter=100, tol=tol, jit=jit, stop_if_linesearch_fails=True).run(x0)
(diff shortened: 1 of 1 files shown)


fabianp:
> @fabianp <https://github.com/fabianp> I noticed you deprecated condition in
> this commit
> <https://github.com/google/jaxopt/commit/7238f760c717e1db534fae3890150b0bfed82a7d>,
> what was the reason for this?
>

The idea was that the options passed with "condition" could now be passed
through the stepsize , since this can be a callable now

fabianp:
the idea is that having a callable for a step-size could be used not just for simple pre-defined schedules, but also for line-searches like (say) a custom Armijo line-search (correct me @vroulet  if I'm wrong)

vroulet:
> the idea is that having a callable for a step-size could be used not just for simple pre-defined schedules, but also for line-searches like (say) a custom Armijo line-search (correct me @vroulet if I'm wrong)

The idea was to avoid making a list of potential linesearches with string options and let the user define his/her own linesearch, which would always be of type IterativeLineSearch. That would have let us avoid adding the arguments of the linesearch inside the solver, but yes it was not finished...

PR #565, opened 2024-01-04
Add .editorconfig

NeilGirdhar:
This file helps non-Googler contributors who have their editors set up
differently for different projects.  Most modern editors will
automatically use the Google style when they see this file:
* (most importantly) 2 character indent size,
* 79 character maximum line length,
* spaces instead of tabs,
* linux line endings, and
* no extraneous whitespace.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,22 @@
+root = true
+
+[*.{py,rst,md,yml}]
+indent_style = space
+end_of_line = lf
+trim_trailing_whitespace = true
+insert_final_newline = true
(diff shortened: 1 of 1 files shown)


vroulet:
Thank you @NeilGirdhar for this, but actually jaxopt was not following the internal Google style. I'll close this.

NeilGirdhar:
> Thank you @NeilGirdhar for this, but actually jaxopt was not following the internal Google style. I'll close this.

Why not edit this to match your style?

PR #568, opened 2024-01-04
Drop Python 3.8

NeilGirdhar:
Jax no longer supports 3.8 (and will drop 3.9 in July according to their [deprecation policy](https://jax.readthedocs.io/en/latest/deprecation.html)).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


mblondel:
Can you also remove it from [here](https://github.com/google/jaxopt/blob/main/.github/workflows/tests.yml)?

NeilGirdhar:
Done :smile:

mblondel (approved):

PR #574, opened 2024-01-25
Fix tests

vroulet:
Tests appeared broken in the main branch because of updates in scikit-learn and scipy.
This fixes several issues by (i) letting scipy run with default precision when testing bfgs/lbfgs, (ii) testing optimum in function values for the lbfgs tests on binary logistic regression, (iii) fixing the argument for the isotonic pav test.


fabianp (approved):
looks good to me, thanks!

Issue #579, opened 2024-02-06
`LevenbergMarquardt` implementation does not accept PyTree parameters


Joshuaalbert:
# Description
The `LevenbergMarquardt` implementation does not accept PyTree parameters, giving `TypeError: primal and tangent arguments to jax.jvp must have the same tree structure` at levenberg_marquardt.py, line 534.

# MVCE

```python
from dataclasses import dataclass
from typing import Literal, NamedTuple, Tuple

import jaxopt
from jax import numpy as jnp


class CalibrationParams(NamedTuple):
    gains_real: jnp.ndarray  # [source, time, ant, chan, 2, 2]
    gains_imag: jnp.ndarray  # [source, time, ant, chan, 2, 2]


class CalibrationData(NamedTuple):
    gains_real: jnp.ndarray  # [source, time, ant, chan, 2, 2]
    gains_imag: jnp.ndarray  # [source, time, ant, chan, 2, 2]


@dataclass(eq=False)
class Calibration:
    convention: Literal['fourier', 'casa'] = 'casa'
    dtype: jnp.dtype = jnp.complex64
    chunksize: int = 1
    unroll: int = 1

    def _residual_fun(self, params: CalibrationParams, data: CalibrationData) -> jnp.ndarray:
        residuals = jnp.concatenate([
            (params.gains_real - data.gains_real).ravel(),
            (params.gains_imag - data.gains_imag).ravel()
        ])
        return residuals

    @property
    def float_dtype(self):
        # Given self.dtype is complex, find float dtype
        return jnp.real(jnp.zeros((), dtype=self.dtype)).dtype

    def get_init_params(self, shape) -> CalibrationParams:
        """
        Get initial parameters.

        Args:
            shape: shape of gains_real and gains_imag

        Returns:
            initial parameters
        """
        return CalibrationParams(
            gains_real=jnp.ones(shape, self.float_dtype),
            gains_imag=jnp.zeros(shape, self.float_dtype)
        )

    def solve(self, init_params: CalibrationParams, data: CalibrationData) -> Tuple[CalibrationParams, jaxopt.OptStep]:
        solver = jaxopt.LevenbergMarquardt(
            residual_fun=self._residual_fun,
            maxiter=100,
            jit=True,
            unroll=False,
            materialize_jac=False,
            geodesic=False,
            implicit_diff=False
        )
        opt_result = solver.run(init_params=init_params, data=data)
        params = opt_result.params
        return params, opt_result


if __name__ == '__main__':
[...]

Issue #505, opened 2023-08-22
LevenbergMarquardt do not seems to work with non-flat input. 


bolducke:
GaussNewton is working as intended with Pytree. I would expect the same for LM. Instead, I had to flatten the array to make it properly works. 

![image](https://github.com/google/jaxopt/assets/26026137/31e9e0de-f7e6-405d-b0c7-18ce08fb317c)

The errors appear at line 445. The error comes from the fact that the pytree of params and vec do not match.

nickmcgreivy:
@amir-saadat I was going to report this as well. I wrote a super simple test case to demonstrate, though I'm not sure its what you're looking for.

```from jax import config
config.update("jax_enable_x64", True)
import jax.numpy as jnp
import jaxopt

M = 5
params = jnp.zeros((2,M))
params = params.at[0].set(jnp.arange(M) * 1.0)
params = params.at[1].set(jnp.arange(M)**2 * 1.0)
params_dict = {'A': jnp.arange(M) * 1.0, 'B': jnp.arange(M)**2 * 1.0}

def F(params):
	return jnp.asarray([jnp.sum(params[0]), jnp.sum(params[0] * params[1]**2)])

def F_dict(params):
	return jnp.asarray([jnp.sum(params['A']), jnp.sum(params['A'] * params['B']**2)])

def optimize_F_gn(params, F):
	gn = jaxopt.GaussNewton(residual_fun=F)
	return gn.run(params).params

def optimize_F_lm(params, F):
	gn = jaxopt.LevenbergMarquardt(residual_fun=F)
	return gn.run(params).params

print(params)
print(optimize_F_gn(params, F)) 
print(optimize_F_gn(params_dict, F_dict))
print(optimize_F_lm(params, F)) # fails
print(optimize_F_lm(params_dict, F_dict)) # fails
```

PR #587, opened 2024-03-31
LevenbergMarquardt and pytrees

gbruno16:
I propose to replace the JAX NumPy operations with the corresponding ones in `tree_utils` to address issues #505 and #579. Now, the snippet in issue #505 appears to run correctly, both with and without geodesic acceleration (using the solver `solve_cg`).
However, QR, LU, and Cholesky still fail since they require the flattened versions of both the Jacobian and parameters.

Regarding the computation of the initial value of the `damping_factor`, using `self.damping_parameter * jnp.max(jtj_diag)` requires materializing the full identity matrix. Perhaps, for large problems like the one in Issue #579, it would be useful to include the option for the user to choose an initial `damping_factor` without calculating `jtj_diag`? (In the same way of the original paper by Marquardt https://www.jstor.org/stable/2098941, p.438)

### DIFF
--- a/jaxopt/_src/levenberg_marquardt.py
+++ b/jaxopt/_src/levenberg_marquardt.py
@@ -35,7 +35,10 @@ from jaxopt._src.linear_solve import solve_cholesky
 from jaxopt._src.linear_solve import solve_inv
 from jaxopt._src.linear_solve import solve_lu
 from jaxopt._src.linear_solve import solve_qr
-from jaxopt._src.tree_util import tree_l2_norm, tree_inf_norm, tree_sub, tree_add, tree_mul, tree_vdot, tree_zeros_like, tree_scalar_mul
+
+from jaxopt._src.tree_util import tree_l2_norm, tree_inf_norm, tree_sub, tree_add
+from jaxopt._src.tree_util import tree_mul, tree_vdot, tree_zeros_like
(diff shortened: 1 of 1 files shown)


vroulet:
Hello @gbruno16,
Good to see you on this repo too!
A few comments:
- We are currently in the process to migrate jaxopt into optax so it may be worth thinking about creating such an optimizer in optax directly. I'd be happy to help in this process but be aware that it'll take some time. In particular optax works with gradient transformation and does not handle solvers as in jaxopt yet. But again I thought about it so we could do it together (ping me by mail if you are interested!).
- If you want to stick with jaxopt's api, I think you can revamp largely this function or even start from scratch with a simpler implementation (it would be helpful for the migration anyway). By simpler implementation, I mean having an implementation that
  - never materializes the jacobian, only works with jvps/vjps (note that once you have e.g. the jvp you can use jax.linear_transpose to get the vjp).
  - uses a cg_sovler that can work directly with linear operators (so no headache of doing an lu etc... and no materialization of the jacobian)
  - no geodesic accelerration to start with (so the code is a bit simpler, although that geodesic acceleration can be add-on later)
   - keep the gain_ratio logic.

gbruno16:
It seems interesting to me! I will send you an email

PR #591, opened 2024-04-25
Fix escape sequences

NeilGirdhar:
Fixes https://github.com/google/jaxopt/issues/598

### DIFF
--- a/jaxopt/_src/osqp.py
+++ b/jaxopt/_src/osqp.py
@@ -330,10 +330,11 @@ class BoxOSQP(base.IterativeSolver):
 
   Dual Eq   variables: :math:`y`
 
-  Dual Ineq variables: :math:`\mu, \phi`
+  Dual Ineq variables: :math:`\\mu, \\phi`
 
 
-  ADMM computes :math:`y` at each iteration. :math:`\mu` and :math:`\phi` can be deduced from :math:`y`.
+  ADMM computes :math:`y` at each iteration. :math:`\\mu` and :math:`\\phi` can be deduced from
+  :math:`y`.
 
   Defaults values for hyper-parameters come from: https://github.com/osqp/osqp/blob/master/include/constants.h
 


NeilGirdhar:
@mblondel Any chance we can get this in soon?

NeilGirdhar:
On Python 3.12, this gives:
```
.../.venv/lib/python3.12/site-packages/jaxopt/_src/osqp.py:299: SyntaxWarning: invalid escape sequence '\m'
```

stephen-huan:
@NeilGirdhar it should also be fixed here.

https://github.com/google/jaxopt/blob/9552f3923d487d2edca3a2c95196b9e77d84dadc/docs/conf.py#L68

+1 for merging, this breaks [GPJax](https://github.com/JaxGaussianProcesses/GPJax)'s tests for me.

NeilGirdhar:
> @NeilGirdhar it should also be fixed here.
> 
> https://github.com/google/jaxopt/blob/9552f3923d487d2edca3a2c95196b9e77d84dadc/docs/conf.py#L68
> 
> +1 for merging, this breaks [GPJax](https://github.com/JaxGaussianProcesses/GPJax)'s tests for me.

Do you mind doing that in separate PR?  I don't use jaxopt anymore and I don't want to deal with the added back-and-forth :/  I think that one should a raw string if it's being passed to re, but I don't want to take the time to figure it out :)

stephen-huan:
A raw string is a Python syntactic construct to create normal strings, but the result is just a normal string once it's passed to a regex function (that is, there's no such thing as a "raw string" type as far as I'm aware). What's shown above is just a normal string, and the same warning applies. (I found this error automatically with [pyright](https://github.com/microsoft/pyright)). Compare

```python
>>> r"_test\.py"
'_test\\.py'
>>> type(r"_test\.py")
<class 'str'>
>>> "_test\.py"
<stdin>:1: SyntaxWarning: invalid escape sequence '\.'
'_test\\.py'
>>> "_test\.py" == "_test\\.py" == r"_test\.py"
<stdin>:1: SyntaxWarning: invalid escape sequence '\.'
True
```

This is minor though, since `conf.py` is some sort of internal documentation generator anyways.

NeilGirdhar:
Yes, I know.  The reason it should be an r-string is because conventionally (according to linters like Ruff), r-strings should be passed as the match pattern of an re-match regardless of whether or not they take advantage of any of differences.  So, your first way of writing it is the conventional way that it should be written rather than the equivalent non-r-string.

stephen-huan:
Which ruff [rule](https://docs.astral.sh/ruff/rules/) are you referring to?

```python
import re

re.match("_test\\.py", "_test\\.py")
```

```console
ruff check test.py
All checks passed!
```

Whether it's `r"_test\.py"` or `"_test\\.py"` is immaterial to me since they're the same. It just can't be `"_test\\.py"`.

NeilGirdhar:
[RUF039](https://docs.astral.sh/ruff/rules/unraw-re-pattern/)

> Whether it's r"_test\.py" or "_test\\.py" is immaterial to me since they're the same. 

Respectfully disagree that it's "immaterial".

tmct:
Hi - please could this either be merged or could we work out an alternative to remove the warning?

For example, we could use raw docstrings instead - I've done this in #625, and checked that the resulting string is equal in a notebook

jaxopt/_src/osqp.py
```diff
@@ -330,10 +330,11 @@ class BoxOSQP(base.IterativeSolver):
 
   Dual Eq   variables: :math:`y`
 
-  Dual Ineq variables: :math:`\mu, \phi`
+  Dual Ineq variables: :math:`\\mu, \\phi`
```
mblondel, line 333: Why not defining the doc-string using `r""" [...] """` instead?

mblondel: Ah sorry, I hadn't seen #625.

vroulet:
Thanks @NeilGirdhar for looking into this. #625 fixed it so I'll close this

NeilGirdhar:
> Thanks @NeilGirdhar for looking into this. #625 fixed it so I'll close this

Great, thanks for sorting this out!

PR #597, opened 2024-05-17
Fix `BoxOSQP` when configured with `fun` and used with pytree-based APIs

diegoferigo:
The combined usage of [`Pytree of matrices API`](https://jaxopt.github.io/stable/quadratic_programming.html#pytree-of-matrices-api) and [`Quadratic function API`](https://jaxopt.github.io/stable/quadratic_programming.html#quadratic-function-api) is currently broken due to a missing operation performed on generic pytrees. This PR updates the operation to be performed with the existing pytree helpers.

### DIFF
--- a/jaxopt/_src/osqp.py
+++ b/jaxopt/_src/osqp.py
@@ -31,7 +31,7 @@ from jax.tree_util import tree_reduce
 from jaxopt._src import base
 from jaxopt._src import implicit_diff as idf
 from jaxopt._src.cond import cond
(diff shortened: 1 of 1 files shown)


diegoferigo:
Failures in CI seem unrelated to the changes of this PR. The [`osqp_test.py`](https://github.com/google/jaxopt/blob/4a50198711e53ed0c25f2be5394825092c2427db/tests/osqp_test.py), only test affected by the proposed changes, is passing.

vroulet (approved):
Thanks for doing this @diegoferigo!

PR #604, opened 2024-07-23
fix lasso with scalar l1reg

BalzaniEdoardo:
First of all, thanks for the very nice package.

In this PR I am generalizing the behavior of `jaxopt.prox.prox_lasso` to handle scalar input when jit-compiled. 

In particular the following used to result in an exception.

```python
import numpy as np
from jaxopt import prox
import jax

rng = np.random.RandomState(0)
x = (rng.rand(20) * 2 - 1, rng.rand(20) * 2 - 1)
jax.jit(prox.prox_lasso)(x, 0.5)
```

Best regards,
Edoardo

### DIFF
--- a/jaxopt/_src/prox.py
+++ b/jaxopt/_src/prox.py
@@ -70,7 +70,7 @@ def prox_lasso(x: Any,
   if l1reg is None:
     l1reg = 1.0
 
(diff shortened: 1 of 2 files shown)


vroulet (approved):
Thanks for doing this

PR #641, opened 2026-06-04
Loosen lbfgsb_test Rosenbrock tolerance to 1.5e-3

seantalts:
Update tolerance - a recent XLA:CPU change to strength reduction of batched vector vector products made this fail due to fp noise.


seantalts:
@vroulet could you take a look? Small test-tolerance bump — a recent XLA:CPU strength-reduction change to batched vector-vector products introduced fp noise that pushes the Rosenbrock value just over the old `1e-3` bound.

vroulet:
The library is no longer maintained, so failing test do not matter I believe. Why do you ask?

PR #644, opened 2026-06-14
docs: add OptStep API reference

### LINKED ISSUE #251: Confusing return type in many solvers
For example, in LBFGS

image

Reading the Return type, it would seem like this returns an object of type OptStep, while in reality, it returns a tuple, of which the second argument is of type OptStep

Haihan-Jiang:
## Summary

- Adds `jaxopt.OptStep` to the API autosummary so solver return types can link to a generated reference page.
- Adds a short docstring for the `OptStep` container fields.

Fixes #251.

## Testing

- `python3.12 -m compileall -q jaxopt/_src/base.py`
- `git diff --check`
- Temporary Python 3.12 venv smoke test with `jax`/`jaxlib`: imported `jaxopt.OptStep` and `jaxopt.base.OptStep`, verified they are the same object, and constructed an `OptStep`.

I did not run the full Sphinx docs build locally because `docs/requirements.txt` installs the full docs stack, including TensorFlow/Flax, for this small autosummary/docstring change.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1,6 +1,14 @@
 API at a glance
 ===============
 
+Base containers
+---------------
+
+.. autosummary::
+  :toctree: _autosummary
+
+    jaxopt.OptStep
+
 Optimization
 ------------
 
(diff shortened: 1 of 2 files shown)


Haihan-Jiang:
This is ready for review from my side. The visible checks are green; could someone take a look when convenient?

Issue #570, opened 2024-01-16
OSQP crashing on unexpected params


Illviljan:
I'm trying to move over some qp code to jaxopt, but I'm struggling to understand the cryptic errors that appears to only happen in the jaxopt implementation. I've tried with other packages and these params work with those implementations.

Here's a minimal example:
```python
import numpy as np
import jax.numpy as jnp
from jaxopt import OSQP

from qpsolvers import solve_qp


def to_numpy(*args):
    return tuple(np.asarray(v) for v in args)


P_ = jnp.array([[576.0]])
q_ = jnp.array([-216.0])
G_ = jnp.array([[-1.0]])
h_ = jnp.array([2.0])
A_ = jnp.array([[]], dtype=float).T
b_ = jnp.array([], dtype=float)

x = solve_qp(*to_numpy(P_, q_, G_, h_, A_, b_), solver="osqp")  # works

qp = OSQP()
deltas = qp.run(
    params_obj=(P_, q_),
    params_eq=(A_, b_),
    params_ineq=(G_, h_),
).params.primal  # Crashes with cryptic error.
# TypeError: dot_general requires contracting dimensions to have the same shape, got (1,) and (2,).

# jax-0.4.23 jaxlib-0.4.23 jaxopt-0.8.3 ml-dtypes-0.3.2 opt-einsum-3.3.0
```

Algue-Rythme:
Hi Illviljan

Sorry for the cryptic error message. The erros come from the fact that the matrix `A_ = jnp.array([[]], dtype=float).T` is not a valid linear operator.  If you don't need equality constraints you just nee to pass `None` to `params_eq`:

```python
P_ = jnp.array([[576.0]])
q_ = jnp.array([-216.0])
G_ = jnp.array([[-1.0]])
h_ = jnp.array([2.0])
# A_ = jnp.array([[]], dtype=float).T
# b_ = jnp.array([], dtype=float)

qp = OSQP()
deltas = qp.run(
    params_obj=(P_, q_),
    params_eq=None,   # CHANGE HERE.
    params_ineq=(G_, h_),
).params.primal
```

Similarly, if you don't need inequality constraints just pass `None` to params_ineq.  Thank you for your message, I just came to the realization that I forgot to document this functionnality.

Illviljan:
Thank you, a quite simple fix. I maybe just need to continue with all constraints active in my larger project.

[...]

PR #646, opened 2026-06-14
docs: clarify OSQP empty constraints

Haihan-Jiang:
Addresses #570.

## What
- Clarify that missing OSQP equality or inequality constraint families should be passed as `None`, not as empty constraint arrays.
- Document that at least one constraint family must be provided for `OSQP`; fully unconstrained QPs should use the unconstrained QP path instead.
- Mirror that guidance in the `OSQP.run` docstring so the generated API docs include it.

## Testing
- `git diff --check`
- `python3.12 -m compileall -q jaxopt/_src/osqp.py`
- Temporary venv smoke test importing `OSQP` and checking the updated `run` docstring
- Temporary venv smoke test solving a small QP with `params_eq=None` and `params_ineq=(G, h)`

The full Sphinx docs build was not run locally because the docs requirements install the full documentation stack, including TensorFlow/Flax.

### DIFF
--- a/docs/quadratic_programming.rst
+++ b/docs/quadratic_programming.rst
@@ -195,6 +195,9 @@ Example::
 
 It is also possible to specify only equality constraints or only inequality
 constraints by setting ``params_eq`` or ``params_ineq`` to ``None``.
+Use ``None`` for a missing constraint family rather than empty arrays, such as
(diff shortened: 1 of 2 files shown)


Haihan-Jiang:
This is ready for review from my side. The visible checks are green; could someone take a look when convenient?

